PR #3, opened 2018-01-28
 1-27-2018 trees-api EDIT + 1

iremaka:
the main change is associated with the parameter signature accepted by
the way you PUT (edit) a tree record. Prior to this, the PUT (edit)
needed all the parameters there in the body for the tree record to be
there - and this may not be desirable.
Now, rather than requiring all parameters, it is supported to PUT for
just some of the parameters. For instance, to change only the latitude
and nothing else is now supported - this can now be done by passing in
only id of the record, and the new latitude, and it will work.

ALSO: a new async series loop to build the query string too not just the
values. Prior to this, while async.eachSeries was used to build the
array of values to pass into the pgsql query, the iterator to build the
array of values to be passed into pgsql was still being constructed with
a forEach iterator. forEach acts like Object.keys in that any usage of
forEach is less likely guarantee the ordering - and moreover forEach is
synchronous and may be a blocking operation. Instead, this usage of
forEach was replaced with an async loop eachOfSeries. In general instead
of using any forEach loops in node.js, prefer using an async function
from the async utility library - e.g. instead of forEach, use
async.eachSeries or async.eachOfSeries

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,6 +17,13 @@ app.use(bodyParser.urlencoded({ extended: false })); // parse application/x-www-
 app.use(bodyParser.json()); // parse application/json
 app.set('view engine','html');
 
+
+const noop = function(){}
+
+const isNotEmpty = function(val){
+    return (val!==null && val!==undefined);
+}
+
 const pool = new Pool({
 
   connectionString: conn.connectionString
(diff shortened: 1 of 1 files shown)


ZavenArra:
Good commentary on this commit!   +1

PR #4, opened 2018-02-16
FILTER BY USER 2-15-2018 +1

iremaka:
FILTER BY USER 2-15-2018 +1

the reading of all points (trees) switched to use crudpool and also to
use the client.query convention.
Moreover the implementation of filtering the points by the user_id.
Zoom was at a static value of 0.01. Changed for now to match the zoom
query parameter.
NOTE that for now, zoom < 14 and user_id cannot be used in conjunction
at the moment.

NOTE: In a demo environment, if all components are on same server and
using nginx,
there are some more changes to nginx config needed for req.params to
work properly.
The nginx config must have the rewrite to preserve the querystring, see
nginx-2-15-2018_1.txt.
Note that dev.treetracker.org may already have this done correctly prior
to this in the nginx configuration.


ZavenArra:
Thank you!!!

PR #15, opened 2018-06-26
Initial docker code

mckornfield:
Not initially working, but I believe it is a config issue.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+node_modules
+npm-debug.log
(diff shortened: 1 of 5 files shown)


mckornfield:
@deepwinter Should be ready for review. Got it working locally now

thiagodemellobueno:
@mckornfield - thanks for this, I'll try and give it a spin this evening and report back.

mckornfield:
Yep no problem... I can work on it more if need be as well.

mckornfield:
@thiagodemellobueno I would test it again, got it running this time by removing the scripts from the docker ignore.

PR #39, opened 2019-03-17
Implement router

gloriaLVela:
Implement the router


ZavenArra:
Hi @gloriaLVela we don't need to rename the service file or create a new one named 'treetracker-admin'.   The repository now contains both the admin frontend and the service for the api, the file that is currently in place is for the api, and is correctly named.

PR #40, opened 2019-03-17
Add column to mark inactive

### LINKED ISSUE #5: Delete record using boolean 'active' field rather than actually deleting.

quorth0n:
Adds a column to mark trees as inactive. Fixes #5 (I think?)

Some notes:
* VSCode kept making a bunch of stylistic changes to code I hadn't touched on save. I think integrating something like prettier with eslint would solve this, or this could have been done already through a different tool and my VSCode config wasn't picking it up. Either way let me know :slightly_smiling_face: 
* I don't have any Redux experience (I only have about 1 yr experience with React, and I've just used hooks/context API to manage state), and I was trying to return the status of the `PATCH` request but couldn't get it to work. I left the dispatcher (check the TODO) in there for someone more senior than me to observe in case you'd like some status checking to be done before merging, otherwise I can just scrap it. Currently after "deleting" an instance it just calls `getTreesAsync` again to refresh.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -4,784 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@babel/code-frame": {
-      "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
-      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
-      "requires": {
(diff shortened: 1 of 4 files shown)


quorth0n:
Screencap:
image

thiagodemellobueno:
@whirish  - we're using a simplified redux called rematch:
See here https://github.com/rematch/rematch/blob/master/docs/api.md#reducers

Some things get abstract and weird, but I think it forces certain patterns, and once the abstractions are grocked, it does making any one piece of functionality pretty easy to implement. Given we're a distributed team with a bunch of contributors, I went for an approach that forces certain conventions / patterns on us, so every bit isnt magically unique.

What you're looking for is probably "effects" .... trees offer an example of how that works with some async bits.

I maybe able to take a look later in the week, but have to attend to other works.

PR #46, opened 2019-03-30
removed special component categorization folders

misterinterrupt:
@deepwinter


misterinterrupt:
got some random formatting in there, but i just want those folders gone

ZavenArra:
Runs fine!

Issue #43, opened 2019-03-20
add imagescrubber actions to imagescrubber


thiagodemellobueno:
We have a husk component for ImageScrubberActions.
At first it may just have:
1. A count of items selected
2. A "delete" button (deactivate tree)
3. Possibly a "we just marked X trees inactive. <undo>" option

ZavenArra:
Tracking this elsewhere

Issue #42, opened 2019-03-20
Add pagination to image scrubber


thiagodemellobueno:
https://github.com/szmslab/material-ui-flat-pagination

PR #47, opened 2019-03-31
Image scrubbing Infinite Scroll

bittricky:
This PR includes changes for issues #43 & #42:
- [react infinite](https://github.com/seatgeek/react-infinite) module for `infinite scroll` on image scrubber
- action for `approving` and `rejecting` tree image


client/src/models/imageScrubber.js
```diff
@@ -83,13 +96,16 @@ const imageScrubber = {
         this.receiveLocation(null, { id: payload.id, address: 'cached' })
       }
     },
-    async markTreeInactive(id) {
+    async toggleTreeActive(id, bool) {
```
mckornfield: maybe instead of bool, `isActive`?

client/src/models/imageScrubber.js
```diff
@@ -48,6 +48,19 @@ const imageScrubber = {
         return { ...state, byId }
       }
     },
+    receiveStatus(state, payload) {
+      /**
+       NOTE: May need to get access to tree store to get the tree to update in the TreeTable - 03/31/19
+       // state.trees.data.forEach(x => {
```
mckornfield: Not sure if you have to implement this, but might be better with an `indexOf( x => x.id === payload.id)` and then just `state.tress.data[index].active = payload.active`

client/src/components/AppFrame/AppFrame.js
```diff
@@ -10,7 +10,8 @@ import Typography from '@material-ui/core/Typography'
 import Divider from '@material-ui/core/Divider'
 import IconButton from '@material-ui/core/IconButton'
 import MenuIcon from '@material-ui/icons/Menu'
-//
+
+import Infinite from 'react-infinite'
```
mckornfield, line 14: do you need to commit this in the `package.json` and `package-lock.json`?

client/src/components/AppFrame/AppFrame.js
```diff
@@ -92,7 +93,9 @@ const AppFrame = (props) => ({
       )
     } else if(currentView === 'imageScrubber') {
       tabContents = (
-        <ImageScrubber />
+        <Infinite containerHeight={200} elementHeight={40}>
```
mckornfield: I think it's probably worth following how the file does it, by adding styles into the styles object

client/src/components/ImageScrubber/ImageScrubber.js
```diff
@@ -51,6 +51,13 @@ class ImageScrubber extends Component {
     this.props.getTreesWithImagesAsync(payload)
   }
 
+  onStatusToggle = (e, id, bool) => {
+    const { toggleTreeActive } = this.props;
+    e.stopPropagation();
+    /** True Active || False Inactive **/
```
mckornfield: if you rename the var, might not be necessary as a comment.

mckornfield (commented):
Some minor style comments, also not sure if you need to add the library (didn't see it when I searched the repo)

client/src/models/imageScrubber.js
ZavenArra: i agree, let's use descriptive variable names.

client/src/models/imageScrubber.js
bittricky: not sure either, but if its updating the tree object in `imageScrubber` should the property `active` be updated on the same tree object in `trees` state as well for consistency?

I'll push up the style fix and leave it commented, for the future.

client/src/models/imageScrubber.js
bittricky: 👍

client/src/models/imageScrubber.js
mckornfield: If it affects the UI (like there's a red box or something) then it should probably be updated for synchronicity's sake. On the other hand, since you hard code the values to either activate or deactivate, you don't have to worry about app state vs. database state for updating.

It's probably fine if you aren't going to show anything as different though

client/src/components/AppFrame/AppFrame.js
bittricky: yup, sorry. Temporary moment of insanity

thiagodemellobueno:
The issue with infinite scroll is that it makes it hard to go to a specific page intentionally, and often (without some robust routing) you lose your place on reload.

This may not matter in this use case, but a UX consideration nonetheless.

This either prioritizes some for of state saving (i still thing we really should implement user accounts as soon as practiceable) and deeplinking/routing.

client/src/components/AppFrame/AppFrame.js
bittricky: Adding these rules to the styles object breaks the module, as the `containerHeight` and `elementHeight` module properties are required and break. Since the styles aren't available when the `react-infinite` module is loaded into the view.

I've wrapped the styles in a separate object to make available and follow the same pattern in commit e2768d9

mckornfield: yep as soon as I suggested this, I also thought it might not work at all hehe (been doing more RN development recently, but in that world you just pass a fat `styles` object) instead of props like this

client/src/components/AppFrame/AppFrame.js
```diff
@@ -92,7 +98,9 @@ const AppFrame = (props) => ({
       )
     } else if(currentView === 'imageScrubber') {
       tabContents = (
-        <ImageScrubber />
+        <Infinite containerHeight={scroll.containerHeight} elementHeight={scroll.elementHeight}>
```
mckornfield, line 101: Not necessary, but out of curiosity can you pass it with the spread right? `<Infinite {...scroll}>

mckornfield (approved):
👍

PR #50, opened 2019-04-05
Basic date range

bittricky:
This pull request includes the following changes:
- documentation update for `server` setup
- bash command to install `node_modules` in server directory
- reused `TreeImageCard` for `ImageScrubber`
- date range button to sort the list of cards by `id` and `timeCreated`


client/src/components/TreeImageCard/TreeImageCard.js
```diff
@@ -0,0 +1,113 @@
...
+})
+
+
+class TreeImageCard extends Component {
+
+  constructor(props) {
+    super(props)
+  }
+
+  onStatusToggle = (e, id, isActive) => {
```
misterinterrupt, line 46: @deepwinter it is un-deleted over here

misterinterrupt:
@bittricky i noticed before that you had a duplication of the treecard component, one in its own file, one in the imagescrubber container component.

this looks like you have remedied this, is the component structure where you were aiming to get it now?

client/src/components/TreeImageCard/TreeImageCard.js
ZavenArra: I guess I am wondering if this is the right pattern, including the toggle callback in the TreeCard?   I know we also toggle active from the list view.

client/src/components/ImageScrubber/ImageScrubber.js
```diff
@@ -149,9 +131,8 @@ const mapState = state => {
 const mapDispatch = (dispatch) => ({
   getTreesWithImagesAsync: ({ page, rowsPerPage, order, orderBy }) => dispatch.trees.getTreesWithImagesAsync({ page: page, rowsPerPage: rowsPerPage, order: order, orderBy: orderBy }),
   getLocationName: (id, lat, lon) => dispatch.imageScrubber.getLocationName({ id: id, latitude: lat, longitude: lon }),
-  toggleTreeActive: (id) => dispatch.imageScrubber.toggleTreeActive(id),
```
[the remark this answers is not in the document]
thiagodemellobueno: This is vestigeous right now, but may come up if we want to show the tree name, though it's fine to remove from now, since we can theoretically duplicate it from the treetable view.

client/src/components/TreeImageCard/TreeImageCard.js
thiagodemellobueno: That file is not referenced anywhere.

misterinterrupt: > 
> 
> That file is not referenced anywhere.

okay, will leave this alone until you have the PR where you want it

client/src/components/ImageScrubber/ImageScrubber.js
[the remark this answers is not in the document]
bittricky: @deepwinter The toggle actions were on the card itself which was a duplicate of the code in `TreeImageCard.js`. Moved the toggle actions into the `TreeImageCard`. To reuse that existing file since the `tree` object is being used to build the image card.

bittricky:
> @bittricky i noticed before that you had a duplication of the treecard component, one in its own file, one in the imagescrubber container component.
> 
> this looks like you have remedied this, is the component structure where you were aiming to get it now?

@thiagodemellobueno started the `TreeImageCard` found in `commit 032763f` for that file, just trying to reuse what was already there from within `../components` for the date range.

client/src/components/ImageScrubber/ImageScrubber.js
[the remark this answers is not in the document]
misterinterrupt: (((fyi, it is not supposed to be a toggle since it is two different properties. one sets active to false and the other sets approved to true, i intend to change that.)))

misterinterrupt:
I love the reuse idea, 
I loaded this branch 
navigated to image scrubber
tried the date sort, it made things disappear.
tried the id sort and it sorts by id and switches the start and finished.
navigated to the tree page, sorting worked as expected for each column.

if the date thing isnt going to work in imagescrubber, then lets remove it.
if it is supposed to work, lets make that happen in the PR that adds the button. sorry if that's obvious and this is still in progress)) :D

bittricky:
@misterinterrupt Here is a gif of what the output should be when running the code from the branch `basic-date-range` with the latest commit from c43bd6a

basic-date-range

I will review your comments and make those changes.

misterinterrupt:
awesome, yeah, that's what I see.
what is going on with the date button there? why do the cards disappear?

bittricky:
@misterinterrupt The date button sorts the list by `id` and `timeCreated` in ascending or descending order. I don't have a definitive answer as to why the cards are not rendering. I can take deeper look as to why its not rendering on the page when its being populated in the redux store.

client/src/components/ImageScrubber/ImageScrubber.js
```diff
@@ -123,11 +123,9 @@ class ImageScrubber extends Component {
             >
           <div className={classes.wrapper}>
             {this.props.treesArray.map(tree => {
-                if (tree.imageUrl) {
```
bittricky: @misterinterrupt Here was the cause of for issue with the `TreeImageCards` not appearing. Many of the cards that are being return by the api do not have value for the `imageUrl` property on each `tree`. 

This condition prevents it from loading onto the page.

client/src/components/ImageScrubber/ImageScrubber.js
```diff
@@ -145,27 +143,27 @@ ImageScrubber.propTypes = {
...
-  byId: PropTypes.object.isRequired,
+  byId: PropTypes.object,
 }
 
 const mapState = state => {
   const keys = Object.keys(state.trees.data)
   return {
     treesArray: keys.map(id => ({
-      ...state.trees.data[id]
+      ...state.imageScrubber.data[id]
```
bittricky, line 153: Switched from using `models/trees` redux store to `models/imageScrubber` for the this page. A lot of the code is similar to the ones used in `models/trees` redux store. 

Since two pages are using `trees`data, it could be worth it to look into having a global store object that is passed to the reducers. This way we could start grafting the store based on reducers.

what do you think @misterinterrupt @thiagodemellobueno @deepwinter?

ref:
[shaping redux store like db](https://hackernoon.com/shape-your-redux-store-like-your-database-98faa4754fd5)

ZavenArra:
Thanks for the fixes and image sort feature!!.  As a next step, a date range picker would be excellent, and is really needed!

PR #53, opened 2019-05-08
remove unneeded depenencies for client

mckornfield:


mckornfield:
@deepwinter I pruned some of the dependencies on the client side to try and get the build to run, these seem to be the bare set needed

ZavenArra:
Requested review from @misterinterrupt.. I'm not sure that we can remove the redux dependency yet, he'll have to comment.

mckornfield:
OK, I just used `depcheck` and it reported these as unused. It also erroneously reported the react-scripts module so I left that in

jameskip (approved):

PR #56, opened 2019-05-15
finite scrolling

misterinterrupt:
so, this works.
didnt want to think about the reusable component design while the feature is needed..

in this PR, you can see that I get a ref to the app frame root component and pass it to the TreeImageScrubber component. all the scroll listening and loading happens here.

### DIFF
--- a/client/src/components/TreeImageScrubber.js
+++ b/client/src/components/TreeImageScrubber.js
@@ -19,7 +19,7 @@ const styles = theme => ({
   wrapper: {
     display: "flex",
     flexWrap: "wrap",
-    padding: "2rem"
+    padding: "2rem 2rem 4rem"
   },
   cardImg: {
     width: "100%",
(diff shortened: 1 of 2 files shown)


client/src/components/TreeImageScrubber.js
```diff
@@ -93,63 +105,96 @@ function TreeImageScrubber({ classes, ...props }) {
...
+        // no more to load!
+        state.noMoreTreeImagesToLoad = true;
+      });
+  };
+
+  const handleScroll = e => {
+    if (
+      state.isLoading ||
+      (scrollContainerRef &&
+        Math.floor(scrollContainerRef.scrollTop) !==
```
misterinterrupt, line 135: floor because these javascript height properties might return decimals..
the behavior is fairly not buggy.. ;)

ZavenArra:
You have succeeded where many have failed!!!!

PR #60, opened 2019-05-17
abusing the api with two order descriptors, what a bad idea

misterinterrupt:
LGTM


ZavenArra:
I am glad that you have seen the error of your ways.

Issue #48, opened 2019-04-03
Formatting / Lint


thiagodemellobueno:
I'm not a fan of airBNB but have used their lint styleguide as formatting enforcement in the past, and though there are some rules that annoy me / seem arbitrary, it's a good starting point.

https://github.com/airbnb/javascript/tree/master/react

dadiorchen:
@Davidezrajay we have defined a simple code style guide in the readme, so I think could close this one, and if anyone has guide suggestions, let's discuss and append to the current guide if need.

Davidezrajay:
@dadiorchen Super awesome. Thanks!

PR #67, opened 2019-05-31
Add lint and begin React Hook implementation

jameskip:
- Add 'standard js' linting
- Conversion of some React Classes to Hooks
- Add max payload to nginx to avoid 413 error when communicating with docker api


jameskip:
closes #48

misterinterrupt:
Hi @jameskip  looks like you are working off of an old version of the repo

PR #68, opened 2019-05-31
Restructure folders

jameskip:


misterinterrupt:
eek, this is also based on the older version.

PR #70, opened 2019-05-31
added auto-loading of tree images 

### LINKED ISSUE #63: no reload after finishing rejects or approvals
once gone through the list of trees there is no reload. Results in an empty screen on admin.treetracker.org

misterinterrupt:
added auto-loading when the number of tree images goes below the page-size (20)
fixes https://github.com/Greenstand/treetracker-admin/issues/63

### DIFF
--- a/client/src/components/TreeImageScrubber.js
+++ b/client/src/components/TreeImageScrubber.js
@@ -63,6 +63,12 @@ function reducer(state, action) {
         isLoading: action.isLoading
       };
       return newState;
+    case "noMoreTreeImages":
+      return {
+        ...state,
+        isLoading: false,
+        moreTreeImagesAvailable: false
+      };
     case "approveTreeImage":
       treeImages = state.treeImages.filter(
         treeImage => treeImage.id !== action.id
@@ -82,15 +88,12 @@ function TreeImageScrubber({ classes, getScrollContainerRef, ...props }) {
   const [state, dispatch] = useReducer(reducer, { ...initialState });
 
   let treeImages = state.treeImages;
-  let scrollContainerRef = null;
+  let scrollContainerRef;
 
   const onApproveTreeImageClick = (e, id) => {
     approveTreeImage(id)
       .then(result => {
         dispatch({ type: "approveTreeImage", id });
-        if (needtoLoadMoreTreeImages()) {
-          // loadMoreTreeImages();
-        }
       })
       .catch(e => {
         // don't change the state if the server couldnt help us
@@ -102,9 +105,6 @@ function TreeImageScrubber({ classes, getScrollContainerRef, ...props }) {
     rejectTreeImage(id)
       .then(result => {
         dispatch({ type: "rejectTreeImage", id });
-        if (needtoLoadMoreTreeImages()) {
-          // loadMoreTreeImages();
-        }
       })
       .catch(e => {
         // don't change the state if the server couldnt help us
@@ -139,7 +139,7 @@ function TreeImageScrubber({ classes, getScrollContainerRef, ...props }) {
       })
       .catch(error => {
         // no more to load!
-        state.moreTreeImagesAvailable = false;
+        dispatch({ type: "noMoreTreeImages" });
       });
   };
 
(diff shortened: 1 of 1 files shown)


client/src/components/TreeImageScrubber.js
```diff
@@ -48,8 +48,8 @@ const initialState = {
   treeImages: [],
   isLoading: false,
   pagesLoaded: -1,
-  noMoreTreeImagesToLoad: false,
-  pageSize: 30
+  moreTreeImagesAvailable: true,
+  pageSize: 20
```
misterinterrupt, line 52: makes the page a bit faster when less are loaded in

client/src/components/TreeImageScrubber.js
```diff
@@ -48,8 +48,8 @@ const initialState = {
   treeImages: [],
   isLoading: false,
   pagesLoaded: -1,
-  noMoreTreeImagesToLoad: false,
-  pageSize: 30
+  moreTreeImagesAvailable: true,
```
misterinterrupt, line 51: less inverted state variable names means less confusion

client/src/components/TreeImageScrubber.js
```diff
@@ -142,24 +156,22 @@ function TreeImageScrubber({ classes, getScrollContainerRef, ...props }) {
...
-    if (scrollContainerRef) {
-      scrollContainerRef.addEventListener("scroll", handleScroll);
-    }
     return () => {
       if (scrollContainerRef) {
         scrollContainerRef.removeEventListener("scroll", handleScroll);
       }
     };
-  }, []);
+  }, [state]);
```
misterinterrupt, line 174: possibly weird, but more specific state change checks seems to make things slower.. we'll see

misterinterrupt:
@deepwinter @jameskip

client/src/components/TreeImageScrubber.js
```diff
@@ -109,23 +116,30 @@ function TreeImageScrubber({ classes, getScrollContainerRef, ...props }) {
...
-        dispatch({ type: "loadMoreTreeImages", treeImages: result });
+        dispatch({
+          type: "loadMoreTreeImages",
+          treeImages: result,
+          isLoading: false
+        });
       })
       .catch(error => {
         // no more to load!
-        state.noMoreTreeImagesToLoad = true;
```
mckornfield, line 128: why does this not use `dispatch`?

client/src/components/TreeImageScrubber.js
```diff
@@ -142,24 +156,22 @@ function TreeImageScrubber({ classes, getScrollContainerRef, ...props }) {
     loadMoreTreeImages();
   };
 
+  scrollContainerRef = getScrollContainerRef() ? getScrollContainerRef() : null;
```
mckornfield: This is a weird idiom (I don't think you wrote this). does `getScrollContainerRef()` not return null? Does it instead return undefined?

mckornfield (commented):
A few comments, mostly unsure about the commented out part

client/src/components/TreeImageScrubber.js
mckornfield: why is this possibly weird?

misterinterrupt: It's not weird on second thought. In the context of using redux, it might be a bad idea to have one component check state that is relevant to all component state. but here, it seems just right

client/src/components/TreeImageScrubber.js
```diff
@@ -87,6 +88,9 @@ function TreeImageScrubber({ classes, getScrollContainerRef, ...props }) {
     approveTreeImage(id)
       .then(result => {
         dispatch({ type: "approveTreeImage", id });
+        if (needtoLoadMoreTreeImages()) {
+          // loadMoreTreeImages();
```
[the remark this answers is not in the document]
misterinterrupt: this is a remnant of how i first tried to do the check for having no images. 
I will delete, since the check is now done whenever state changes, not just in approve/reject.

misterinterrupt:
@mckornfield thanks a ton for the review! super valuable comments and the code is now reflecting your wisdoms :p

client/src/components/TreeImageScrubber.js
```diff
@@ -81,7 +88,7 @@ function TreeImageScrubber({ classes, getScrollContainerRef, ...props }) {
   const [state, dispatch] = useReducer(reducer, { ...initialState });
 
   let treeImages = state.treeImages;
-  let scrollContainerRef = null;
+  let scrollContainerRef;
```
mckornfield, line 91: 👍

mckornfield (approved):
Looks good

PR #92, opened 2019-07-22
installed Storybook; add Filter component and filter function;

dadiorchen:
@deepwinter I made some progress on filter function, it is not a perfect version, some UI is not good yet but It does work, if you need this function urgently ( filter by ID, status, created time), please accept this pull. 

about the filter:
Trees can be filtered by ID, status, created time:
Screen Shot 2019-07-22 at 11 12 02 AM

problem:
* UI is not so good, and didn't follow the spec too much. And because the table component is difficult to adjust the width, so I keep filter open.
* Material-UI date picker is hard to use, should be replaced in the future.
* I am not sure about the spec of "status", I just send the text directly to Backloop, such as: not a tree. 

Bonus:
Installed storybook, and I add some stuff maybe it is useful for others and designer:
Screen Shot 2019-07-22 at 11 12 24 AM

To run the Storybook:
```
cd client
npm run storybook
```


ZavenArra:
Thanks!!

PR #101, opened 2019-10-07
new date picker

wadaemmanuel:
adjusted date picker in the filter component


client/src/components/Filter.js
```diff
@@ -253,20 +257,10 @@ function Filter(props){
 				justify='space-between'
 			>
 				<Grid item>
-					<OutlinedInput
-						type='date'
-						className={classes.dateInput}
-						value={dateStart}
-						onChange={handleDateStartChange}
-					/>
+					<DateSelector/>
```
dadiorchen: @wadaemmanuel how you sure it works in this way?

dadiorchen (commented):
@wadaemmanuel It looks like this version is not a working version?  I don't think it will work well for the filter to let it set the date correctly, because you didn't pass the date down to the date picker component and retrieve the date back from it

Issue #111, opened 2019-11-10
Clearing the filter


sebastiangaertner:
Clearing the filter is currently not possible.

sjaymoon15:
PR #114 for this issue has been merged to master branch.

PR #114, opened 2019-12-05
Clearing the filter (Issue #111)

nickharrisdev:
Implemented a new button in the filter form which clears all the fields and resets the results shown. 

Also made some linter-related, code-style changes throughout the file. The primary changes related to clearing the filter are located lines 69-80 and 131-137. 

Screen Shot 2019-12-04 at 5 58 09 PM


dadiorchen:
@nickharriscodes as we talked about this morning, we need you to pull the newest version of code from the master branch, because we made some change in codes. Sorry about the chaos. 
After updating the code, please put back your code again.
FYI, I use below tip to update fork, do you have any better way? please let me know.
https://gist.github.com/CristinaSolana/1885435

nickharrisdev:
@dadiorchen I was able to redo my changes in the new component. Let me know what you think and if I need to make any more changes.

dadiorchen:
hey, great, @nickharriscodes , just need adjust some UI, these two buttons are too close, give some margin would be better
Screen Shot 2019-12-20 at 8 55 47 AM

sjaymoon15:
I see there's a related test file called `Filter.test.js`. 
We might need to start suggesting updating related test as much as possible. 

Just a thought.

dadiorchen:
@sjaymoon15 can not agree more :) 
my opinion about this is that we should let  our model pass a fully test as much as possible and because the component is not so easy to code unit test, and maybe require/force people to write component layer test is too much

nickharrisdev:
@sjaymoon15 & @dadiorchen I agree, writing tests is definitely good practice. There isn't much in the README about testing. Are we using Jest? And how do you run the tests? I don't know a lot about writing testing suites, but I'm definitely interested in learning more. Maybe the testing portion could be broken out into a separate ticket?

Other than the testing stuff, is there anything else to add? If not, I think this ticket is done.

dadiorchen:
``@nickharriscodes yes, we use Jest, and we just follow the React convention to do that, run ```npm test``` you can run all the tests.
Maybe this simple tutorial is helpful,  https://github.com/Greenstand/treetracker-admin/blob/master/rematchTutorial.md
at the bottom, there is a unit test example.
https://github.com/Greenstand/treetracker-admin/blob/master/rematchTutorial.md#unit-test-example

dadiorchen:
for this ticket, please refine the UI, currently, those two buttons are too close. put some margin/gap between them.

nickharrisdev:
@dadiorchen I added 5px margin between the two buttons on my last commit. Will that suffice?

dadiorchen:
I missed the notification that you had committed that change, yes, it's OK now, merge to master

PR #124, opened 2020-01-11
Update Date Picker API

NickSiena:
Updated Date Picker API with new Material UI. Updated 1/11/2020 by Nick Siena.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,20 +1,12 @@
-{
-  "name": "treetracker-admin",
-  "version": "1.2.1",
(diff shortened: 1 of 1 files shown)


sjaymoon15:
There might have been a mistake during making this PR because it looks like it's overwriting the `package.json` file entirely ?

NickSiena:
@sjaymoon15 Dadiorchen told me to delete and replace the old code with the new code. That's exactly what I did. Thanks.

ZavenArra:
Package.json was replaced with html file, which is a clear error in this PR

Issue #123, opened 2020-01-06
improve the data picker component


dadiorchen:
Now, we are using the built-in date picker in material-UI,  it 's not good, including bad UX, and bad UI, and prone to misoperation. So we need to replace it with a better one, materialUI recommend an alternative, (https://material-ui.com/components/pickers/#material-ui-pickers) , I thought it looks good, and if you have other suggestion, welcome.
<img width="329" alt="Screen Shot 2019-10-01 at 8 46 10 AM" src="https://user-images.githubusercontent.com/5744708/71789655-4c69fa80-3067-11ea-89a0-5683b2db2c57.png">
<img width="361" alt="Screen Shot 2019-10-01 at 8 46 25 AM" src="https://user-images.githubusercontent.com/5744708/71789661-52f87200-3067-11ea-8c5b-9e6103fd35b7.png">

PR #127, opened 2020-01-15
Update/material ui pickers

sjaymoon15:
Implementation of Issue #123

### DIFF
--- a/client/src/components/Filter.js
+++ b/client/src/components/Filter.js
@@ -48,8 +48,8 @@ const styles = theme => {
 
 function Filter(props) {
   const { classes, filter } = props;
-  const dateStartDefault = '1970-01-01';
-  const dateEndDefault = `${dateformat(Date.now(), 'yyyy-mm-dd')}`;
+  const dateStartDefault = null;
+  const dateEndDefault = null;
   const [treeId, setTreeId] = useState(filter.treeId);
   const [planterId, setPlanterId] = useState(filter.planterId);
   const [deviceId, setDeviceId] = useState(filter.deviceId);
@@ -65,13 +65,15 @@ function Filter(props) {
   const [dateEnd, setDateEnd] = useState(filter.dateEnd || dateEndDefault);
 
   const handleDateStartChange = date => {
-    const formattedDate = dateformat(date, 'yyyy-mm-dd');
-    setDateStart(formattedDate);
+    setDateStart(date);
   };
 
   const handleDateEndChange = date => {
-    const formattedDate = dateformat(date, 'yyyy-mm-dd');
-    setDateEnd(formattedDate);
+    setDateEnd(date);
+  };
+
+  const formatDate = date => {
+    return dateformat(date, 'yyyy-mm-dd');
   };
 
   function handleClear() {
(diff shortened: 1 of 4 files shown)


client/src/components/Filter.js
```diff
@@ -1,306 +1,293 @@
...
+    setDateStart(formattedDate);
+  };
 
-	function handleDateEndChange(e){
-		setDateEnd(e.target.value || dateEndDefault)
-	}
+  const handleDateEndChange = date => {
+    const formattedDate = dateformat(date, 'yyyy-mm-dd');
+    setDateEnd(formattedDate);
+  };
```
sjaymoon15: Changed `handleDateStartChange` and `handleDateEndChange` for the date picker change.

client/src/components/Filter.js
```diff
@@ -1,306 +1,293 @@
...
+            format='MM/dd/yyyy'
+            margin='normal'
+            id='end-date-picker'
+            label='End Date'
+            value={dateEnd}
+            onChange={handleDateEndChange}
+            KeyboardButtonProps={{
+              'aria-label': 'change date'
+            }}
+            className={classes.dateInput}
```
sjaymoon15, line 280: Replaced the current date picker with material ui date picker. `OutlinedInput`s were replaced with `KeyboardDatePicker`

client/src/components/Filter.js
```diff
@@ -1,306 +1,293 @@
...
+import MenuItem from '@material-ui/core/MenuItem';
+import FilterModel from '../models/Filter';
+import dateformat from 'dateformat';
+import GSInputLabel from './common/InputLabel';
+import classNames from 'classnames';
+import DateFnsUtils from '@date-io/date-fns';
+import {
+  MuiPickersUtilsProvider,
+  KeyboardDatePicker
+} from '@material-ui/pickers';
```
sjaymoon15, line 19: Libraries added for this update.

sjaymoon15:
Thanks for testing and checking for different behaviors. Will look into this soon !

sjaymoon15:
1. because the new component can not set year directly( I'm correct?), so we must be sure we can input the date string in the input field, now, it does not work, I get error when I delete chars
=> You are correct.
=> And This has been added in the last commit

2.you might notice that I set the date (1970-now) by default, actually this is bad implement, I have to do so because the old component is ugly, I must set some date to avoid some kind of error( I forgot the detail), so if you can remove the default date ( that means if users do not set date proactively, then date is not a filter condition) , that would be wonderful!
=> This has been added in the last commit

3. when I click a date on the component, why it can not select the date and disappear? yes, the date was selected, but the component can't disappear after clicking.
I'm using Chrome 79
=> In order to make the calendar disappear after clicking a date, you need to click anywhere outside of the calendar (or simply press ESC key). This seems to be the default behavior of the date picker library. (https://material-ui.com/components/pickers/#material-ui-pickers)


Would love to get some feedback on these changes.
Especially for number 2 change, in the Filter.js file (client/src/models/Filter.js),rn It does not seem to handle the case when both start date and end date are falsy.

dadiorchen:
I will check the number 2 on my pc, I thought you are correct, Filter.js maybe do not support this change

about the UX of this component, if found a demo on materialUI
Untitled_ Jan 18 2020 9_07 AM
, maybe we can follow this one?  If we use this demo, I thought it will benefit both number 1, and number 3, support year selection and click 'OK' to finish the operation.\

thanks for your efficient work!

sjaymoon15:
Update has been pushed to this PR. 
I replaced the Date picker inline with the Date picker dialog which is the one you shared in the last comment. I like that option to also pick the year as well.  @dadiorchen

dadiorchen:
thanks! I think it's good enough to merge to reop, although there might be tiny bugs in this function, let's solve it in future when we get some bugs, ( mostly it's because we changed the logical, remove the default date as a filter condition)

dadiorchen:
solved #123

Issue #113, opened 2019-11-18
"Reject All" for bulk reject


sebastiangaertner:
Additional to the "Approve All" there must be a "Reject All" batch button.

qdcheng1:
Can I take this?

Davidezrajay:
@david-mike This is still a valid ticket. Thanks it would be great to have!

qdcheng1:
This ticket could be closed?

PR #128, opened 2020-01-22
Reject All

qdcheng1:
Add Reject All feature for bulk reject. Modify the verity state to support reject all.

### DIFF
--- a/client/src/components/TreeImageScrubber.js
+++ b/client/src/components/TreeImageScrubber.js
@@ -328,6 +328,39 @@ const TreeImageScrubber = ({ getScrollContainerRef, ...props }) => {
                     onClick={async () => {
                       if (
                         window.confirm(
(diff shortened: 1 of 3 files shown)


Davidezrajay:
#113

dadiorchen:
David, I tried your code, found some problem, it seems like when rejecting all operation, it would led to some disorder status, look below gif, when I rejected a tree, then tried to undo this operation, then I got a tree list with wrong order, looks like the order direction has been reversed, and I thought I met some other wrong case, try this: ( select a range, approve all, then undo it, then select another range, then reject all, check the result)
Untitled_ Jan 24 2020 10_40 AM

dadiorchen:
@david-mike

qdcheng1:
For the first case, it did cause wrong order. I found the same thing happen to Approve All as well. Since they call the same undo method. I need to fix the undo method. I did not get the other cases. Could you explain them in details?

PR #160, opened 2020-04-04
Trees to verify counter V2

### LINKED ISSUE #133: Show "trees to verify" counter
Show "# trees to verify" 
This number is the number of trees have not been given any tags.

gracewylu:
Second attempt at trees to verify counter:
- [x] API call to grab number of unverified trees
- [x] updates count when trees get approved/rejected

- resolves https://github.com/Greenstand/treetracker-admin/issues/133

Screen Shot 2020-04-04 at 6 57 34 PM

### DIFF
--- a/client/src/api/treeTrackerApi.js
+++ b/client/src/api/treeTrackerApi.js
@@ -81,5 +81,9 @@ export default {
     })
       .then(handleResponse)
       .catch(handleError);
+  },
+  getUnverifiedTreeCount() {
+    const query = `${baseUrl}/trees/count?where[approved]=false&where[active]=true`;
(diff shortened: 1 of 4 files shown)


dadiorchen:
@gracewylu very nice! It is reaching the satisfaction almost, just one thing, I hope you can add a test to cover your code added into the model file, I mean the stuff you added into the client/src/models/verity.js file, can you do that? We have some examples in the test file, client/src/models/verity.test.js, ask me if you run into problem.

gracewylu:
@dadiorchen I tried adding a basic test case but let me know if I need to add more!

PR #176, opened 2020-05-06
fix docker for development 

### LINKED ISSUE #174: Docker is broken...
Out docker for development is broken, can not start up our client, seeking one who is familiar with docker to help us out.

gracewylu:
- fixes https://github.com/Greenstand/treetracker-admin/issues/174

Docker initially wasn't working for me but I was able to find this github issue (https://github.com/facebook/create-react-app/issues/8688) that listed the following:

> Adding `stdin_open: true` to the docker-compose file solved the issue for me

It seems like a recent update to `react-scripts` has caused Docker to fail for many others and I think because we updated `package.json` in our latest merged in code it caused us to update `react-scripts` as well. 

> Adding `stdin_open: true` or `tty: true` to the docker-compose file works.
> 
> stdin_open stands for interactive and tty means terminal. If you do a quick google search you find this:
> Screen Shot 2020-03-24 at 11 22 29
> 
> [7e6d6cd](https://github.com/facebook/create-react-app/commit/7e6d6cd05f3054723c8b015c813e13761659759e) clearly indicates a check for interactivity.
> 
> docker run by default allocates tty. So that's why that works.
> https://docs.docker.com/compose/reference/run/
> 
> Docker's documentation says: "For interactive processes (like a shell), you must use -i -t together in order to allocate a tty for the container process."
> 
> https://docs.docker.com/engine/reference/run/#foreground

Would like for others to test before & after to see if this PR fixes the docker issue for them as well!

### DIFF
--- a/dev/compose/docker-compose.yml
+++ b/dev/compose/docker-compose.yml
@@ -13,6 +13,7 @@ services:
      - "3001:3000"
     volumes:
      - ../../client:/opt/admin-web
+    stdin_open: true
   treetracker-admin-api:
     container_name: "treetracker-admin-api"
     image: "treetracker-admin-api:latest"


dadiorchen:
@gracewylu Awsome! It works, You made huge help!

Issue #175, opened 2020-05-04
Add check box to selected images


Davidezrajay:
When an image is selected we should add a check box to it as per the designs. 
 Desired state:
![Screenshot_20200504-111305](https://user-images.githubusercontent.com/25363572/81005199-09901780-8dfa-11ea-926b-547b980c60bc.png)

nmcharlton:
Mind if I take this one?
Can you point me at some sample data to populate the admin panel?
Is there a full set of design material, including style guidance?

dadiorchen:
Hi, @nmcharlton that's great you can deal with this one. Please first install the dev env locally then you can find there are some data in it already. 
We are using MaterialUI, so you need to follow the rule of them.

nmcharlton:
Thanks @dadiorchen.
To get things going I needed to change `variables.js` to point at my local API container as well as adding a valid `treetracker.datasource.json`. I'm using [a version I found on Slack](https://greenstand.slack.com/files/U73TYBBAB/FL2U6N7QE/treetracker.datasource.json) - is this an appropriate data source for development?

nmcharlton:
Are the checkboxes just intended a visual cue to indicate that an image is selected (so can be hidden when not selected) or should a user be able to toggle individual checkboxes as a means of selecting/deselecting images?

nmcharlton:
> Are the checkboxes just intended a visual cue to indicate that an image is selected (so can be hidden when not selected) or should a user be able to toggle individual checkboxes as a means of selecting/deselecting images?

Is this the sort of behaviour expected from this interaction?

![OneDrive_multi-select](https://user-images.githubusercontent.com/5558838/81827819-0c44e800-9531-11ea-8871-709b97e22e28.gif)

dadiorchen:
@nmcharlton Hey, you are right, it's reasonable, just show the checkbox when selected. 

> Thanks @dadiorchen.
> To get things going I needed to change `variables.js` to point at my local API container as well as adding a valid `treetracker.datasource.json`. I'm using [a version I found on Slack](https://greenstand.slack.com/files/U73TYBBAB/FL2U6N7QE/treetracker.datasource.json) - is this an appropriate data source for development?

Yes, it is.

dadiorchen:
Why do you need to change variable to run it? I thought we can use it directly? Whan kind of problem you met?

nmcharlton:
[...]

PR #187, opened 2020-05-14
Added check marks to selected images

nmcharlton:
Fixes Issue #175 

Since no checkbox functionality is required, only the display of check marks, there is no need for an interactive checkbox component.


nmcharlton:
Screenshot:
checked_selection

dadiorchen:
Great job!

Issue #167, opened 2020-04-28
Sort the list of species. 


dadiorchen:
In version 2, the list of species, in the right-side-panel, and in the filter, the order is messy, should sort the list by alphabet.
<img width="246" alt="Screen Shot 2020-04-28 at 5 10 24 PM" src="https://user-images.githubusercontent.com/5744708/80469473-2f896800-8973-11ea-8230-b01c9aa0d10b.png">
<img width="371" alt="Screen Shot 2020-04-28 at 5 10 08 PM" src="https://user-images.githubusercontent.com/5744708/80469498-3912d000-8973-11ea-8149-43fae265c5db.png">

nmcharlton:
I'll take a look at this.

PR #190, opened 2020-05-14
Sorted species lists alphabetically in filter and tags

nmcharlton:
Addresses issue #167 

A copy of the species list is sorted using `localCompare()` on the species names at the point of rendering the list.

### DIFF
--- a/client/src/components/FilterTop.js
+++ b/client/src/components/FilterTop.js
@@ -253,7 +253,9 @@ function Filter(props) {
                 setSpeciesId(e.target.value)
               }
             >
-              {[{id:0,name:'all'}, ...props.speciesState.speciesList].map(species => (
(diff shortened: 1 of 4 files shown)


dadiorchen (changes requested):
nmcharlton, can you do this in the model? https://github.com/Greenstand/treetracker-admin/blob/master/client/src/models/species.js 
I thought that would be more reasonable to do so. No need to do the same thing at every place where you use a special list.

nmcharlton:
> nmcharlton, can you do this in the model? https://github.com/Greenstand/treetracker-admin/blob/master/client/src/models/species.js
> I thought that would be more reasonable to do so. No need to do the same thing at every place where you use a special list.

That makes sense.

dadiorchen (approved):
Hey, wonderful you checked the tests and wrote new one!

PR #191, opened 2020-05-16
Fix Menu Bug [75%]

### LINKED ISSUE #166: Fix the menu bug in new version.
In the new version, the menu bar is broken, can't click the menu item to jump the correct pages, needs to be fixed. (branch v2.1)
Screen Shot 2020-04-24 at 7 40 04 PM

### LINKED ISSUE #179: "Trees" menu item not working in production
left menu item "trees" seems to not respond.

gracewylu:
- resolves https://github.com/Greenstand/treetracker-admin/issues/166
- resolves https://github.com/Greenstand/treetracker-admin/issues/179

Managed to fix the Menu so that one can toggle between the `Verify` and `Trees` but right now the Filter toggle button doesn't work (since it is specific to the `<TreeImageScrubber>` component). Trying to figure out a way to show/get working the filter toggle only on the menu for the Verify page. Any advice on solving this from those more experienced with React would be greatly appreciated!

Screen Shot 2020-05-15 at 11 17 32 PM
Screen Shot 2020-05-15 at 11 17 47 PM

### DIFF
--- a/client/src/components/MainFrame.js
+++ b/client/src/components/MainFrame.js
@@ -3,10 +3,13 @@
  */
 import React		from 'react';
 import { makeStyles } from '@material-ui/core/styles';
+import { useTheme }   from '@material-ui/styles';
+import AppBar from '@material-ui/core/AppBar';
 import Menu		from './common/Menu';
 import Filter		from './Filter';
 import Grid		from '@material-ui/core/Grid';
 import Box		from '@material-ui/core/Box';
+import MenuIcon from '@material-ui/icons/Menu';
 import Table		from '@material-ui/core/Table';
 import TableHead		from '@material-ui/core/TableHead';
 import TableRow		from '@material-ui/core/TableRow';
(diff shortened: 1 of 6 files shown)


client/src/components/MainFrame.js
```diff
@@ -93,30 +108,84 @@ function SimpleTable() {
...
+                <Grid item>
+                  <IconButton
+                    onClick={handleFilterClick}
+                  >
+                    <IconFilter />
+                  </IconButton>
+                </Grid>
+              </Grid>
+            </Grid>
+            {/*isFilterShown &&
```
dadiorchen, line 161: The filter was removed? How would it work then?

dadiorchen (commented):
In my previous thought, this issue is a hassle, that's because our app layout frame is kind of unusual, I am afraid we need to do some ugly job to let things happen.

client/src/components/MainFrame.js
gracewylu: Yeah that's the problem I'm having right now since it looks like we want the Menu bar on all the pages but only the filter button when we're viewing TreeImages. Was hoping for some advice maybe on how to navigate this issue! We could potentially create 2 separate top navbars, one for the TreeImages and another for everything else. 

For now I had just commented out the Filter part so that the page doesn't break when testing it out. Let me know if you'd like for me to keep digging into a solution!

PR #202, opened 2020-05-19
Made tree image grid responsive

### LINKED ISSUE #184: Adjust the tree picture size to fit the container.
Currently, every line there are 3 trees displaying, but the size is not so appropriate, need to adjust the size to be as big as possible to fill the space.
Screen Shot 2020-05-14 at 9 26 35 AM

nmcharlton:
Resolves #184 

- Changed tree image layout to use Material UI Grid with appropriate breakpoints
- Set tree image height relative to card width
- Reduced left and right margins to better make use of space
- Fixed card focus animation

Images below show card sizing at different screen widths.

**800px**

**1280px**

**1440px**

**1920px**


dadiorchen:
This is a good one.

PR #207, opened 2020-05-22
Replace the table component. 

### LINKED ISSUE #183: Replace the table component.
Currently, we are using table component, mui-datatables, but we found there are many problems and limitations for using it. Like it's hard to change the appearance of it, and customizing something. 
So we decide to change back to use the original MaterialUI table. 
This ticket is about removing the component and changing to MaterialUI table component. 
Any suggestion is welcome.

lukasbicus:
closes #183

Material UI table looks like this:
Screenshot 2020-05-22 at 16 25 23

### DIFF
--- a/client/src/components/TreeTable.js
+++ b/client/src/components/TreeTable.js
@@ -1,16 +1,13 @@
-import React, { Component } from 'react';
-import compose from 'recompose/compose';
-import { connect } from 'react-redux';
-import { withStyles } from '@material-ui/core/styles';
-import MUIDataTable from 'mui-datatables';
-import dateformat		from 'dateformat';
-import Filter, {FILTER_WIDTH}		from './Filter';
-import FilterModel		from '../models/Filter';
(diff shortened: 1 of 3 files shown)


client/src/components/TreeTable.js
```diff
@@ -20,195 +17,172 @@ const styles = () => ({
...
+            ))}
+          </TableBody>
+        </Table>
         <Drawer
           anchor="right"
           open={this.state.detailsPane}
-          onClose={event => this.toggleDrawer(event, undefined)}
+          onClose={(event) => this.toggleDrawer(event, undefined)}
         >
           <TreeDetails tree={tree} />
```
dadiorchen, line 173: This just can show the first page, and can't do pagination and others.

dadiorchen (commented):
This PR is unsatisfied, it removed the Material-table, but didn't use appropriate equivalent function to replace it, the new version just can show the first page, and can't go to next and other pages.

PR #215, opened 2020-05-25
Added pagination to the Verify page

### LINKED ISSUE #104: Add "select number of images" for Verify Page
Currently, the Verify Page auto loads images when user reaches the bottom of the page. 
Screenshot (49)

The desired function is the same as the function displayed on the trees page as shown below. 

 Screenshot (50)_LI

This is the Verify page that should be modified.
Screenshot (52)

nmcharlton:
Resolves #104 

- Pagination component above and below images
- 12, 24, 48 or 96 images per page
- Tree count updated to reflect filter used for tree images
- Only loads more trees if there are more available and not enough to populate the current page

- Placeholder cards while images are loading


nmcharlton:
I'd welcome thoughts on possible tests that could be added.

dadiorchen:
It looks good, let's push to test env, and let our tester try it :).

PR #223, opened 2020-05-28
Extracted Navbar into a separate component

### LINKED ISSUE #208: Build a nav bar component.
Currently, we need to copy the navbar on every page: 
Screen Shot 2020-05-23 at 10 18 04 AM
So it would be better to build a component to re-use it. 
The problem is: we have some "special" part in the navbar when displayed on different pages. Like this one, the filter is different from "verify" page. 
Screen Shot 2020-05-23 at 10 18 17 AM
So this component is not just a static one, it should have some dynamic content in it.

nmcharlton:
Resolves #208 

This change has preserved the existing appearance of the navigation bar on the Planters and Verify pages and added it to the top of the Trees page.
This means that the navigation menu is accessible from the Trees page.

A few additional steps were required to get cypress working for component testing.


dadiorchen (approved):
Great, I am happy to see you are using Cypress, :)

Issue #243, opened 2020-06-02
User System, a secure loophole. 


dadiorchen:
Currently, we use JWT to encode the data, now the key for JWT is hardcoded in the repo, we need to separate it to a config file, and remove it from the repo. Just like what we did for the database source.

elfiyang16:
I'd like to take this.

PR #249, opened 2020-06-03
User system/issue 243

elfiyang16:
- add key to a dotenv file that is referenced in the config (where we can put more stuff in it later?)
- will pin the jwt key in the slack admin_panel

https://github.com/Greenstand/treetracker-admin/issues/243#issue-628822333


dadiorchen (approved):
This is a beautiful PR

PR #250, opened 2020-06-04
Separated env files into dev, test and prod

### LINKED ISSUE #172: Map link direct as per configured
Admin panel: map link only directs to treetracker.org, it needs to be configurable to use the map for the environment it's installed on and configured for.

- test.treetracker.org
- Test.treetracker,org, 
- dev.treetracker.org

nmcharlton:
Resolves #172 

The `.env` file in the client has been complemented with `.env.development`, `.env.test`, `.env.staging` and `.env.production` to set environment variables specific to each environment, including the webmap domain.

The webmaps used in each environment are given below, together with the corresponding `npm` commands.

### dev.treetracker.org

- development: `npm start`

### test.treetracker.org

- test (local): `npm test`
- staging: `npm run build:staging`

### treetracker.org

- production: `npm run build`

Automated build and deploy scripts may need to be updated to accommodate the staging environment.


dadiorchen:
I saw the .env.production,  but when do we use it? shouldn't we have a script like: build:production ?
Then I thought it is more consistent? npm run build:production

nmcharlton:
> I saw the .env.production, but when do we use it? shouldn't we have a script like: build:production ?
> Then I thought it is more consistent? npm run build:production

When running `npm run build`, production is the default environment (as dictated by npm and react-scripts), meaning `NODE_ENV` is set to `production` and `.env.production` is always used.

The idea is that `build:staging` is a special case. `.env.production` is still read, but vars are overridden by those in `.env.staging`.

Issue #235, opened 2020-06-01
Users should be able to change password of themselves. 


dadiorchen:
Click the 'change' button to change user's password. Need to input three form input:
* Current password
* New password.
* Confirm password.

NOTE, this action needs to solve the 401 permission problem for the password change API, needs to allow any user to change the password of themselves, currently, the password change API just allows the user who has 'user manager' permission to do so. 

<img width="1552" alt="Screen Shot 2020-06-01 at 5 41 18 PM" src="https://user-images.githubusercontent.com/5744708/83397219-d4dea280-a42f-11ea-845d-137a65cb7883.png">

elfiyang16:
In order to do this, do you mean to change endpoint from 
      `${process.env.REACT_APP_API_ROOT}/auth/admin_users/${userPassword.id}/password` to 
      `${process.env.REACT_APP_API_ROOT}/admin_users/${userPassword.id}/password`, ?
@dadiorchen

elfiyang16:
Also do we have a design mock for this form or just freestyle 😃 @dadiorchen

elfiyang16:
I will take this

dadiorchen:
Thank you @elfiyang16

PR #252, opened 2020-06-05
User system/user form pwdchange/issue 235

elfiyang16:
https://github.com/Greenstand/treetracker-admin/issues/235#issue-628300485

**create pwd change form** 
- currently show 3 different error message and 1 success message as shown below 
Jun-05-2020 21-47-34
Jun-05-2020 21-23-37
image

image

P.S i comment out line 76 in server/js/auth.js since I'm not sure what it's doing and I can't bypass it....happy to change back if required 
image


dadiorchen:
@elfiyang16  This is great! 
About init(), this one is to avoid we deploy to a new environment and don't have initial data yet, it is kinda ugly but helpful for saving some trouble, for us, can I help you to bypass it? What's the problem with this?

server/src/js/auth.js
```diff
@@ -192,6 +192,23 @@ router.get('/admin_users/', async (req, res, next) => {
   }
 });
 
+router.post('/validate/', async (req, res, next) => {
```
dadiorchen, line 228: It's fine you use this path for the changing-password API

server/src/js/auth.js
```diff
@@ -73,7 +73,7 @@ router.get('/permissions', async function login(req, res, next) {
 router.post('/login', async function login(req, res, next) {
   try {
     //try to init, in case of first visit
-    await init();
+    // await init();
```
dadiorchen: If remove this, we will get some trouble when we deploy the system to a new environment without initial data in DB

elfiyang16:
> @elfiyang16 This is great!
> About init(), this one is to avoid we deploy to a new environment and don't have initial data yet, it is kinda ugly but helpful for saving some trouble, for us, can I help you to bypass it? What's the problem with this?

Oh I see, thanks for clarification. So the problem is when we execute `/login` command for example, `init()` get executed, and when it found there's existing data in the db, it will simply console log`There are accounts in admin user, quit.` and return, so `login` never get executed @dadiorchen

dadiorchen:
Can we uncomment the line of init()? Is there any trouble doing so? @elfiyang16

PR #263, opened 2020-06-12
Use ctrl+click to toggle selection state of tree

### LINKED ISSUE #262: Tree verify: press CTRL/CMD to select trees.
Currently, our multiple-selection rule is:
* click tree to select a single tree.
* press SHIFT and click to select consecutive trees. 
* press CMD/CTRL and click to unselect one single tree.

Now, in the real practice, operators want to select trees one by one and they are consecutive. 
So, I suggest to add: 
* press CMD/CTRL to select one single tree.

This means: if the tree was selected, then, press CMD/CTRL to unselect it. otherwise, select it.

nmcharlton:
Resolves #262 

ctrl-select

### DIFF
--- a/client/src/models/verity.js
+++ b/client/src/models/verity.js
@@ -518,9 +518,15 @@ const verity = {
 					treeImagesSelected,
 				})
 			} else if (isCmd || isCtrl) {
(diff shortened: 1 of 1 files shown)


dadiorchen:
Thanks Nick

PR #264, opened 2020-06-13
feat: only show error when focus

elfiyang16:
As title illustrates
Jun-13-2020 13-26-59


client/src/components/Login.js
```diff
@@ -173,10 +181,18 @@ const Login = (props) => {
             type="password"
             id="password"
             autoComplete="current-password"
+            onFocus={() => {
+              setPasswordFocus(true)
+            }}
+            onBlur={() => {
+              setPasswordFocus(false)
```
dadiorchen: In my opinion, we should not set it back to false, because after user have touched the input, we should show the error message, even though the user blurs the input.

dadiorchen (changes requested):
And when user clicks the submit button, we should set all the focus variables to true, to let user see the problem.
like: 
```
handleSubmit(){
  setNameFocus(true);
  setPasswordFocus(true);
}
```
like this library: https://jaredpalmer.com/formik/docs/overview  they have a 'touched' variable for every input field and set the 'touched' to true once user focused them.

dadiorchen:
Oh, I noticed the helptext is showing all the time. How about using a variable "touched", and use it to control both helptext and error

Issue #234, opened 2020-06-01
User System: need to add 'created time' on the user list, detail, account places.


dadiorchen:
Please add the create time of the user at these places:
<img width="1552" alt="Screen Shot 2020-06-01 at 5 36 35 PM" src="https://user-images.githubusercontent.com/5744708/83396459-97c5e080-a42e-11ea-9bfe-554eeeb7ba09.png">
<img width="1552" alt="Screen Shot 2020-06-01 at 5 37 17 PM" src="https://user-images.githubusercontent.com/5744708/83396478-9d232b00-a42e-11ea-8773-90e3a16c7aeb.png">
<img width="1552" alt="Screen Shot 2020-06-01 at 5 36 54 PM" src="https://user-images.githubusercontent.com/5744708/83396482-9e545800-a42e-11ea-891e-3ea85240791c.png">

Please be aware, when display in "user detail", created time should be uneditable.

ericlvogt:
I can take this one.

PR #265, opened 2020-06-15
Add created time of user

ericlvogt:
#234  as requested:
image
image
image


client/src/components/Users.js
```diff
@@ -394,7 +395,7 @@ function Users(props) {
                   <Table className={classes.table} aria-label="simple table">
                     <TableHead>
                       <TableRow>
-                        <TableCell>Userame</TableCell>
+                        <TableCell>Username</TableCell>
```
dadiorchen, line 398: Thanks :)

dadiorchen (approved):

Issue #229, opened 2020-06-01
User System: need to add 'delete user' function.


dadiorchen:
Need to add 'delete user' function, on the user manager page. 
<img width="1440" alt="Screen Shot 2020-06-01 at 5 18 41 PM" src="https://user-images.githubusercontent.com/5744708/83395000-1e2cf300-a42c-11ea-97e6-0aeb6c074749.png">

ericlvogt:
I can take this one

PR #278, opened 2020-06-22
Issue 229 User Manager Delete

ericlvogt:
#229 Delete functionality added.


server/src/controllers/adminUser.controller.ts
```diff
@@ -83,4 +83,17 @@ export class AdminUserController {
     await this.adminUserRepository.updateById(id, adminUser);
   }
 
+  @del('/admin_user/{id}', {
```
dadiorchen: I thought we have deprecated this controller and just coding in auth.js?

ericlvogt: I did not realize this was deprecated I see now the /api requests are handled by the loopback request handler and the /auth by the express router . Thanks for letting me know. Is the intent to eventually move all API calls from the .ts controllers? Also should the controller adminUser.controller.ts be deleted/commented deprecated to ensure no one else makes the same mistake/uses this controller?

dadiorchen: Yes, @ericlvogt actually it's my fault, I should have removed these stuff away to avoid misunderstanding, could you comment this part? Then I will merge this PR to master.

ericlvogt: No problem @dadiorchen I should've noticed during development. I've commented all API request handlers for this controller but left the admin controller, model, and repository for now.

PR #283, opened 2020-06-24
bypass JWT authentication for API explorer, update readme file

### LINKED ISSUE #244: Our loopback api explorer is broken.
Previously, when developing, we can use Loopback rest API explorer to check/test the API, now, because we added user system, so it is blocked by the 401 auth check, can we find a way to work around it to let people use the explorer to make development easier?
image

EltonGarcia:
Fix #244 

Authentication was blocking the explorer HTTP  requests.

Should we validate the development environment before bypassing the authentication?

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ docker logs -f treetracker-admin-api
 
 Also see _Scripts_ below
 
-The REST API documentation can be viewed and explored by visiting http://localhost:3000/explorer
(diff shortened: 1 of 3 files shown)


dadiorchen:
> Should we validate the development environment before bypassing the authentication?

I think this is a good solution, cuz we also need to deny the request from the public users when we push it on to the production env. So, we can check: if it's on dev env, then bypass the auth check. I think this is will solve the problem perfectly. 
@EltonGarcia can you add it?

Issue #300, opened 2020-07-01
UX issue: It's not clear which tags are saved on tree rejection


nmcharlton:
The *Tags* sidebar on the *Verify* page has a set of tag options, some of which fall under *Approve* and *Reject* headings.

What's not clear is that when you reject a tree, only the `rejection_reason` is recorded.
All other tags (`morphology`, `age` and `species_id`) are discarded, even though the inputs are still enabled and visible.

Perhaps all tag inputs should sit under either *Approve* or *Reject*.

![tags](https://user-images.githubusercontent.com/5558838/86240938-f3a09800-bb99-11ea-8c6f-4d6faa2b3e2e.png)

![reject](https://user-images.githubusercontent.com/5558838/86241276-8c371800-bb9a-11ea-8da4-1a8c2540bbf0.png)

ericlvogt:
I'll handle this. You are right, only rejection reason is recorded on reject I"ll make this clear to the user.

PR #320, opened 2020-07-22
Verify page moved tags under approve

ericlvogt:
Moved tags under approve there should be no confusion now.
image

### DIFF
--- a/client/src/components/TreeImageScrubber.js
+++ b/client/src/components/TreeImageScrubber.js
@@ -586,54 +586,6 @@ function SidePanel(props){
       elevation={11}
     >
       <Grid container direction={'column'} className={classes.sidePanelContainer}>
-        <Grid>
-          <Typography variant='h5' >Tags</Typography>
-        </Grid>
-        <Grid className={`${classes.bottomLine} ${classes.sidePanelItem}`}>
-          <RadioGroup value={morphology} className={classes.radioGroup}>
-            <FormControlLabel 
-              value='seedling' 
-              onClick={() => handleMorphology('seedling')} 
-              control={<Radio/>} 
-              label='Seedling' />
-            <FormControlLabel 
-              value='direct_seedling' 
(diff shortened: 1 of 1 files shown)


dadiorchen:
@Davidezrajay Do you think this change is reasonable?  Comparing what we have currently:

image

Move the tag and species port into the section of 'approve'.

ericlvogt:
Forgot to link the issue before. This PR addresses issue #300 to prevent confusion of what is saved on submit. On reject submit only the reject reason is saved therefore all other tags have been moved under approve.

dadiorchen:
@Davidezrajay what's you thoughts on this?

client/src/components/TreeImageScrubber.js
```diff
@@ -654,34 +606,84 @@ function SidePanel(props){
...
-              <FormControlLabel 
-                onClick={() => handleCaptureApprovalTag('palm')}
-                value='palm' control={<Radio/>} label='Palm' />
-              <FormControlLabel 
-                onClick={() => handleCaptureApprovalTag('timber')}
-                value='timber' control={<Radio/>} label='Timber' />
-            </RadioGroup>
+            <Grid container direction={'column'} className={classes.sidePanelContainer}>
+              <Grid>
+                <Typography variant='h6' >Tags</Typography>
```
nmcharlton, line 611: I think the __Tags__ title should be above the tabs as it applies to both approval and rejection

nmcharlton (dismissed):
Sorry for the long delay in reviewing.

client/src/components/TreeImageScrubber.js
nmcharlton: Scrap that - it's fine as it is.

TANguyen1893:
Any further development should be done in the admin panel client repo: https://github.com/Greenstand/treetracker-admin-client

Issue #318, opened 2020-07-19
Using 'React-Router' to deal with the page navigation. 


dadiorchen:
Currently, we just use React Context to handle page jumping, the menu changing, now with the project becomes bigger and bigger, 'React Router' turns out to be more and more desired for us. So let's use React Router to handle this project's page transition.

elfiyang16:
I will take this

dadiorchen:
:) @elfiyang16 I am glad you are interested in this, @nmcharlton have you started for this? Or let @elfiyang16 take it

nmcharlton:
@dadiorchen I haven’t started this one.

@elfiyang16 It’s all yours 😄

elfiyang16:
Thanks will push it soon.

PR #322, opened 2020-07-22
Feat/react router

elfiyang16:
https://github.com/Greenstand/treetracker-admin/issues/318#issue-661097120

Jul-22-2020 10-55-09


package.json
```diff
@@ -2,7 +2,10 @@
   "name": "treetracker-admin",
   "version": "1.2.1",
   "description": "This portion of the project is to process tree data. TreeTracker's Admin Panel Frontend and RESTful API built with loopback.",
-  "dependencies": {},
+  "dependencies": {
```
dadiorchen: @elfiyang16 Why do you install the router here? It supposes to be under client , right?

PR #362, opened 2020-08-20
Organization

ZavenArra:


client/src/api/treeTrackerApi.js
```diff
@@ -1,6 +1,15 @@
 import { handleResponse, handleError } from "./apiUtils";
 import {session} from "../models/auth";
 
+
+function getOrganization(){
```
ZavenArra: Looks like we are replicating this code block ( I also see it above).  Can we break this out into a function call in a shared file so that we don't make any mistake.

server/src/repositories/trees.repository.ts
```diff
@@ -13,4 +14,18 @@ export class TreesRepository extends DefaultCrudRepository<
...
+    expect(organizationId).number();
+    expect(this).property("execute").defined();
+    const result = await this.execute(`select * from getEntityRelationshipChildren(${organizationId})`, []);
+    return result.map(e => e.entity_id);
+  }
+
+  async getPlanterIdsByOrganizationId(organizationId : Number):Promise<Array<Number>>{
+    expect(organizationId).number();
+    expect(this).property("execute").defined();
+    const result = await this.execute(`select * from planter where person_id in (select entity_id from getEntityRelationshipChildren(${organizationId}))`, []);
```
ZavenArra: If the intention is to get all planter ids within an organization, then this query needs to look up planters by organization_id matching the entities in getEntityRelationshipChildren

server/src/controllers/treesOrganization.controller.ts
```diff
@@ -0,0 +1,185 @@
...
+  })
+  async count(
+    @param.path.number('organizationId') organizationId: number,
+    @param.query.object('where', getWhereSchemaFor(Trees)) where?: Where<Trees>,
+  ): Promise<Count> {
+    const entityIds = await this.treesRepository.getEntityIdsByOrganizationId(organizationId);
+    const planterIds = await this.treesRepository.getPlanterIdsByOrganizationId(organizationId);
+    where = {
+      ...where,
+      or: [
```
ZavenArra, line 46: Good.   We are making one change in the database that will change this soon, but we need this OR for now.  When Mark is finished though we will have a database trigger to set planting_organization_id when trees are created based on the planter's organization, so we don't have an issue if the planter starts planting for a different organization.

ZavenArra (commented):
Great work!  Just a couple of comments

server/src/repositories/trees.repository.ts
dadiorchen: So I need to change 'person_id' to 'organization_id' that's all right?

dadiorchen: @deepwinter

client/src/api/planters.js
```diff
@@ -1,9 +1,13 @@
-import { handleResponse, handleError } from "./apiUtils";
+import { 
+  handleResponse, 
+  handleError,
+  getOrganization,
```
dadiorchen, line 4: Moved duplicated code block to common utils.

server/src/repositories/trees.repository.ts
```diff
@@ -13,4 +14,19 @@ export class TreesRepository extends DefaultCrudRepository<
...
+  async getEntityIdsByOrganizationId(organizationId : Number):Promise<Array<Number>>{
+    expect(organizationId).number();
+    expect(this).property("execute").defined();
+    const result = await this.execute(`select * from getEntityRelationshipChildren(${organizationId})`, []);
+    return result.map(e => e.entity_id);
+  }
+
+  async getPlanterIdsByOrganizationId(organizationId : Number):Promise<Array<Number>>{
+    expect(organizationId).number();
+    const result = await this.execute(`select * from planter where organization_id in (select entity_id from getEntityRelationshipChildren(${organizationId}))`, []);
```
dadiorchen, line 27: Changed to use organization_id

dadiorchen (commented):
@deepwinter made some change according to your opinion.

Issue #368, opened 2020-08-26
Display appropriate message when login failed


dadiorchen:
If user login admin panel with the username which does not exist in the DB, should display message like: "the user name 'xxx' does not exist"
Now, the message is not so clear, and the server response 500 rather than 401. Also maybe we should consider the response spec to let the client know what happened on the server?
<img width="1438" alt="Screen Shot 2020-08-26 at 4 05 58 PM" src="https://user-images.githubusercontent.com/5744708/91278304-a108e400-e7b6-11ea-930f-77a18a0b3c65.png">

alansvits:
I am trying fix it. Intermediate result like this:
![noexistuser](https://user-images.githubusercontent.com/15067440/92134762-98e22180-ee12-11ea-8953-b74ed843c4da.png)

PR #379, opened 2020-09-06
Appropriate message when login failed

alansvits:
#368 Now error 401 is  send if query with username returns 0 rows. This  error is catched.

### DIFF
--- a/client/src/components/Login.js
+++ b/client/src/components/Login.js
@@ -168,11 +168,18 @@ const Login = (props) => {
           setLoading(false)
         }
       } catch (e) {
-        console.error(e)
-        setErrorMessage(
-          'Could not log in. Please check your username and password or contact the admin.'
-        )
-        setLoading(false)
+        console.error("Undefined User error:", e)
+        if (e.response.status === 401) {
+          setErrorMessage(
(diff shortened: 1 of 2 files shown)


client/src/components/Login.js
```diff
@@ -168,11 +168,18 @@ const Login = (props) => {
           setLoading(false)
         }
       } catch (e) {
-        console.error(e)
-        setErrorMessage(
-          'Could not log in. Please check your username and password or contact the admin.'
-        )
-        setLoading(false)
+        console.error("Undefined User error:", e)
+        if (e.response.status === 401) {
```
nmcharlton: This logic is not quite right – 401 is used for all 'failed to log in' responses.
A better option would be to include the error message in the response from the server in all cases (as you have done for the new error).
If no error message is included, we could fall back to the generic message.

nmcharlton (changes requested):

client/src/components/Login.js
alansvits: I think I got it. Thanks for detailed feedback!

nmcharlton (approved):

Issue #375, opened 2020-09-03
Do not send request to server if #username# field is blank at login page 


alansvits:
I think there should be some client check against sending blank username or password fields to server.

alansvits:
I added a few checks. Now button has disable style if either of fields are empty. #380

PR #380, opened 2020-09-06
Do not send request to server if #username# field is blank at login page

alansvits:
#375 added a few checks for empty fields before submitting

### DIFF
--- a/client/src/components/Login.js
+++ b/client/src/components/Login.js
@@ -265,7 +265,7 @@ const Login = (props) => {
             variant="contained"
             color="primary"
             //Disable button if sending request, username or password are empty
-            disabled={loading || (!userName && usernameBlurred) || (!password && passwordBlurred)}
+            disabled={loading || !userName || !password}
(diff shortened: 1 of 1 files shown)


client/src/components/Login.js
```diff
@@ -259,7 +264,8 @@ const Login = (props) => {
             fullWidth
             variant="contained"
             color="primary"
-            disabled={loading}
+            //Disable button if sending request, username or password are empty
+            disabled={loading || (!userName && usernameBlurred) || (!password && passwordBlurred)}
```
nmcharlton: We could simplify this a little and just disable the button if either `userName` or `password` is empty (without checking `blurred`.
I think that would make the user experience cleaner.

alansvits: So log in button state will be disabled when page is loaded, right
Screenshot 2020-09-07 at 10 58 04

nmcharlton (approved):
Perfect 😄

PR #393, opened 2020-09-13
Filter trees on Verify page by tree tag

### LINKED ISSUE #224: Add search by Tags and Species to Tree Verify Page
Ideally we would be able to search by trees so we can get a list/ review specific species.

nmcharlton:
Resolves #224 

User can specify a single text tag by which to filter trees.
This currently only includes tags added in the 'Additional tags' field when verifying a tree.

Screenshot 2020-09-13 at 19 22 50


server/src/controllers/trees.controller.ts
```diff
@@ -16,9 +16,15 @@ import {
   // del,
   requestBody,
 } from '@loopback/rest';
+import {ParameterizedSQL} from 'loopback-connector';
```
dadiorchen, line 19: I thought you are pretty professional now in terms of Loopback, :D

server/src/models/index.ts
```diff
@@ -3,4 +3,4 @@ export * from './planter.model';
 export * from './species.model';
 export * from './adminUser.model';
 export * from './tag.model';
-export * from './treeTag.model';
+export * from './tree-tag.model';
```
dadiorchen: Looks like a little different from the convention of naming. Why we must name the model name with dash?
And I noticed that you change underscore to dash, we have convention for that?

dadiorchen (approved):

server/src/models/index.ts
nmcharlton: This is the style used by the Loopback v4 CLI for multi-word models/endpoints.
I thought it easiest to go along with it, but I'm happy to revert back to the old format for consistency.
What do you think?

nmcharlton: I'll change back to match the style of other filenames and endpoints.
There's nothing worse than inconsistency.

PR #404, opened 2020-09-21
Move the hard-coded policy of permission into the env config files

dadiorchen:
To avoid hard code for the permission policy item, move those items to the environment configuration files, both on the client-side and server-side.
And we'd better check if these configurations were set correctly when the app starts, so it's easy to find the problem if we missed them in different env.


nmcharlton:
@dadiorchen I'm not sure about putting all of these strings in env vars - it seems a little overkill.
Will these strings change in different environments?

PR #412, opened 2020-09-23
Stay on same page on reload

### LINKED ISSUE #398: Router always redirects to /login on page reload
When the user is logged into the Admin Panel and reloads a page, the browser is always routed to `/login` and then to `/` (home).

Redirection should only happen when a user is not logged in (e.g. their session has expired).

Screen Recording 2020-09-18 at 14 39 47

nmcharlton:
Resolves #398 

This PR improves the routing flow of the client:

- When logged in, user reloads a page → same page loads without rerouting
- When not logged in, user attempts to visit a private route (e.g. `/trees`) → user is redirected to `/login`, then to the private route (if they have permission)
- When logged in, user attempts to visit a private route they do not have permission to access → user is redirected to `/` (home)
- 'Remember me' is now ticked by default, since the current behaviour when unticked is to log the user out when they reload the page – this can be addressed separately


client/src/components/Routers.js
```diff
@@ -39,13 +41,18 @@ export default function Routers() {
...
+                    state: { from: location }
+                  }}/>
+                )}
+              />
             </Switch>
           </Grid>
         </Grid>
       </Grid>
     )
-  }, [appContext.routes])
```
BirdTho, line 50: Why change this? Namely, the routes object is memoized so that we don't waste time recomputing the routes if the item doesn't change. If the user changes (via deep equals) so does the routes, but the current implementation may set the same user object a few times, causing the context object to be regenerated. But the pointer to context.routes will be the same.

If the permissions change, the user changes, and routes changes with it, afaik

client/src/components/Users.js
```diff
@@ -152,7 +152,9 @@ function Users(props) {
     }
   }
 
-  React.useEffect(load, [])
```
BirdTho, line 155: Was this so bad?

BirdTho (changes requested):

client/src/components/Users.js
```diff
@@ -413,9 +415,9 @@ function Users(props) {
         </TableCell>
         <TableCell>{user.active ? 'active' : 'inactive'}</TableCell>
         <TableCell>
-          {user.role.map((r, i) => (
-            <Grid key={i}>
-              {permissions.reduce((a, c) => a || (c.id === r ? c : undefined), undefined).roleName}
+          {user.role.map((r, idx) => (
+            <Grid key={`role_${idx}`}>
+              {permissions.reduce((a, c) => a || (c.id === r ? c.roleName : undefined), undefined)}
```
BirdTho, line 420: I noticed that some users have role IDs 5 and 6, which no longer exist as permissions. Some have these roles multiple times. I considered turning permissions into a key/value map or array to make this process faster. map * reduce is O(m * n), but map * lookup is O(m). More of a comment, not a thing that needs done. It also would be nice to have the invalid ones filtered out rather than mapping over those values and generating JSX tags

dadiorchen (approved):

client/src/components/Users.js
nmcharlton: No, sorry. I meant to change this back.

client/src/components/Users.js
nmcharlton: I agree, this certainly needs tidying up at some point.

client/src/components/Routers.js
nmcharlton: The additional check for `appContext.isLoggedIn()` to decide between `/` and `/login` triggered a React dependency warning suggesting `appContext` needed to be included as a dependency (react-hooks/exhaustive-deps).

I assume this is React saying that it can't know what state changes in `appContext` would cause changes in the value returned by `appContext.isLoggedIn()`.

I'll change `isLoggedIn` to be a variable rather than a method and include it directly as a dependency.

client/src/components/Users.js
nmcharlton: Scratch that, it does cause a problem. If the function is passed directly as a parameter like this, it causes a runtime warning in the browser:

```
Warning: An effect function must not return anything besides a function, which is used for clean-up.

It looks like you wrote useEffect(async () => ...) or returned a Promise. Instead, write the async function inside your effect and call it immediately:

useEffect(() => {
  async function fetchData() {
    // You can await here
    const response = await MyAPI.getData(someId);
    // ...
  }
  fetchData();
}, [someId]); // Or [] if effect doesn't need props or state
```

When you navigate away from the User Manager page, React then fails to clean up correctly and throws some nasty errors.

Ideally, the async function would sit inside the scope of the `useEffect` callback, but in this instance it is called in several other places and I wanted to avoid too much rework.

PR #416, opened 2020-09-28
Delete password hash and salt before returning users

### LINKED ISSUE #415: Strip password hash and salt from admin_user query response
The password hash and salt from the `admin_user` table is included in the response from the server.
This is not good; these fields are secret and should always be stripped.

nmcharlton:
Resolves #415


dadiorchen (commented):
That's correct!

PR #421, opened 2020-10-02
fix: bug

### LINKED ISSUE #420: Fix species table not scrollable and no-jumping-to-the-start when selecting rows per page

elfiyang16:
resolve #420

Oct-02-2020 09-47-16

### DIFF
--- a/client/src/components/SpeciesTable.js
+++ b/client/src/components/SpeciesTable.js
@@ -28,6 +28,8 @@ import { withStyles } from '@material-ui/core/styles'
 
 const styles = (theme) => ({
   box: {
+    width: "80%",
+    overflow: 'auto',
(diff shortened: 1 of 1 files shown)


client/src/components/SpeciesTable.js
```diff
@@ -28,6 +28,8 @@ import { withStyles } from '@material-ui/core/styles'
 
 const styles = (theme) => ({
   box: {
+    width: "80%",
```
nmcharlton: This looks a bit odd - there's a gap to the right of the scroll bar.

nmcharlton (changes requested):
Can we make the contents scroll and keep the menu fixed on the left?

PR #383, opened 2020-09-09
#329 Create CI/CD Pipeline

ghost:
**Summary of changes:**

1. Create Build Stage of CI/CD Pipeline.
Notes:
Pushing commits (or doing pull requests) to the server project will trigger server pipeline.
Likewise, pushing commits (or doing pull requests) to the client project will trigger client pipeline.
2. Update package.json of client project with basic `lint` command.
3. Configure Loopback API to use environmental variables for database configuration.
4. Write sample integration test for species controller to utilize environmental variables.

PR #435, opened 2020-10-22
Create GitHub Actions CI pipeline for Client and Server

### LINKED ISSUE #329: Setup Github Actions for CI and pull request.
Setting Github Actions to build the project and run all tests after posting a PR. 
https://docs.github.com/en/actions/language-and-framework-guides/using-nodejs-with-github-actions 
When test failed, there should be a warning on the PR page.

nmcharlton:
Resolves #329 

Largely copied from PR #383  

The server build currently fails since it requires `treetracker.datasource.json` to be included in the build.
We exclude this database configuration file from the repo to avoid publicising credentials.
This should be addressed as a separate issue, most likely by moving these credentials to environment variables.


client/src/components/Home.js
```diff
@@ -62,7 +62,7 @@ import {
 //   </svg>
 // )
 
-const style = (theme) => ({
+const style = () => ({
```
dadiorchen, line 65: So changes like this are for the purpose that we can pass the lint checking? 
I am thinking that maybe the problem like violate the rule of code style and lint warning, is not necessary to check for the code PR commit and deploy, what do you think?

dadiorchen (commented):
I just doubt if checking code format and lint are necessary for committing PR.

And am I correct, if people post PR, we will run the tests including lint and prettier checking?

client/src/components/Home.js
nmcharlton: Although some lint warnings are related to code style, this step is mostly there to catch potential bugs with static analysis. For example, an unused parameter like this might indicate a scoping problem within the function, or where the developer has forgotten to read the input and has used a member variable instead.

The workflow doesn't run `prettier`, just `eslint`. Warnings and errors from `eslint` should be addressed before the code is deployed, even if the appropriate _solution_ is to suppress a warning.

PR #440, opened 2020-10-31
Replace treetracker.datasource.json with DATABASE_URL env var

### LINKED ISSUE #436: Replace treetracker.datasource.json with environment variable
We currently use a json datasource file to tell LoopBack where to look for its database, but the requirement to keep this file out of git for security purposes causes the build process to fail in the GitHub Actions workflow (see PR #435).

Instead, we should use an environment variable to set the database URL and remove the datasource config file.

nmcharlton:
Resolves #436 

The previous mechanism for specifying the database connection (`treetracker.datasource.json`) has been replaced with an environment variable (`DATABASE_URL`) to allow server CI pipeline builds to succeed.

In the development environment, `DATABASE_URL` should be set in `server/.env.development`.
For running tests, the same env var can be set in `server/.env.test`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Treetracker Admin Panel
 
-This portion of the project is to process tree data. TreeTracker's Admin Panel Frontend and RESTful API built with loopback.
+This portion of the project is to process tree data. Treetracker's Admin Panel Frontend and RESTful API built with loopback.
 
 See [Wiki](https://github.com/Greenstand/treetracker-admin-api/wiki) for more info on goals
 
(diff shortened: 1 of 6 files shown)


dadiorchen:
So in the case of CI, the env variable would overwrite the same variable in the .env files, right?

nmcharlton:
> So in the case of CI, the env variable would overwrite the same variable in the .env files, right?

Effectively, yes.
`.env` files in the server are not committed to the repo, so are not available to CI.
Any env vars (such as this one) must be set in the CI workflow.

PR #451, opened 2020-11-15
Allow users to show only trees with no species set

### LINKED ISSUE #428: Add `none` to species search window
This search box is awesome and has pushed us so far forward! 

Add the ability to see which trees have not been tagged. There is the counter to the `all` feature. The default should be a `none` or `not tagged`. 

This gives the use the ability to see all the trees that have no species attached to them.

Screenshot (151)

nmcharlton:
Resolves #428 

Selecting _Not set_ from the Species dropdown in the Verify page filter will show only trees whose species has not been set.

This required switching tree requests to the Loopback JSON object format.

Screenshot 2020-11-15 at 15 45 27

### DIFF
--- a/client/src/api/treeTrackerApi.js
+++ b/client/src/api/treeTrackerApi.js
@@ -10,7 +10,7 @@ export default {
     //the filter model
     filter,
   }) {
-    const where = filter.getBackloopWhere()
+    const where = filter.getWhereObj()
    
     const lbFilter = {
       where,
@@ -122,7 +122,7 @@ export default {
   getTreeCount(filter) {
     const query = `${
       process.env.REACT_APP_API_ROOT
-    }/api/${getOrganization()}trees/count?where=${JSON.stringify(filter.getBackloopWhere())}`
+    }/api/${getOrganization()}trees/count?where=${JSON.stringify(filter.getWhereObj())}`
     console.log(query, session.token)
     return fetch(query, {
       headers: {
--- a/client/src/api/treeTrackerApi.js
+++ b/client/src/api/treeTrackerApi.js
@@ -10,25 +10,29 @@ export default {
     //the filter model
     filter,
   }) {
-    const query =
-      `${process.env.REACT_APP_API_ROOT}/api/${getOrganization()}trees?` +
-      `filter[order]=${orderBy} ${order}&` +
-      `filter[limit]=${rowsPerPage}&` +
-      `filter[skip]=${skip}&` +
-      `filter[fields][imageUrl]=true&` +
-      `filter[fields][lat]=true&` +
-      `filter[fields][lon]=true&` +
-      `filter[fields][id]=true&` +
-      `filter[fields][timeCreated]=true&` +
-      `filter[fields][timeUpdated]=true&` +
-      `filter[fields][active]=true&` +
-      `filter[fields][approved]=true&` +
-      `filter[fields][planterId]=true&` +
-      `filter[fields][deviceId]=true&` +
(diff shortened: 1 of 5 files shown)


client/src/api/treeTrackerApi.js
```diff
@@ -10,25 +10,29 @@ export default {
...
+        timeUpdated: true,
+        active: true,
+        approved: true,
+        planterId: true,
+        deviceId: true,
+        planterIdentifier: true,
+      },
+    }
+    
+    const query = `${process.env.REACT_APP_API_ROOT}/api/${getOrganization()}trees?filter=${JSON.stringify(lbFilter)}`
```
mckornfield, line 35: ah this is much better

client/src/models/Filter.test.js
```diff
@@ -1,41 +1,35 @@
...
+		filter.planterIdentifier = '1'
 	})
 
-	it('getBackloopString() should be: ', () => {
-		console.log(filter.getBackloopString())
-		expect(filter.getBackloopString()).toEqual(expect.stringContaining('&filter[where][id]=10'))
-		expect(filter.getBackloopString()).toEqual(expect.not.stringContaining('&filter[where][status]=planted'))
-		expect(filter.getBackloopString()).toEqual(expect.stringContaining('&filter[where][timeCreated][between]=2019-07-25&filter[where][timeCreated][between]=2019-07-30'))
+	it('getWhereObj() should be: ', () => {
+		console.log(filter.getWhereObj())
```
mckornfield: should we remove this `console.log`?

client/src/models/trees.js
```diff
@@ -139,14 +139,24 @@ const trees = {
...
-          process.env.REACT_APP_API_ROOT
-        }/api/${getOrganization()}trees?filter[order]=${orderBy} ${order}&filter[limit]=${rowsPerPage}&filter[skip]=${
-          page * rowsPerPage
-        }&filter[fields][id]=true&filter[fields][timeCreated]=true&filter[fields][status]=true` +
-        `&filter[fields][planterId]=true&filter[fields][treeTags]=true&filter[where][active]=true` +
-        (filter ? filter.getBackloopString() : '')
+      const where = filter ? filter.getWhereObj() : {}
+   
+      const lbFilter = JSON.stringify({
+        where: {...where, active: true},
```
mckornfield, line 145: so this always filters for active, no matter what's passed in?

client/src/api/treeTrackerApi.js
mckornfield: it seems weird there's no slash between `${getOrganization()}` and `trees`, but I guess that's the way it was?

mckornfield (approved):
overall lg, might want to squash your commits into one, esp since the 3rd and fourth involve undoing on another

client/src/api/treeTrackerApi.js
nmcharlton: Yep, I agree it looks a bit odd.
If the user has an organization, the slash is appended to the org path.
Otherwise, there's no slash so that the call hits the `trees` endpoint.

client/src/models/Filter.test.js
nmcharlton: Agreed.

client/src/models/trees.js
nmcharlton: That's right. We're using `active=true` to mean _not deleted_.
Perhaps this isn't the best place to be setting it though; it might make more sense if it's returned as part of `where` from `filter.getWhereObj()

PR #459, opened 2020-11-28
Add planter registration date to Planter Details

### LINKED ISSUE #390: Planter registration date needs to be shown
We need to add the registration data of a user and ideally see if a planter has been onboarded.  For now a date in the user list would be great. 

Screenshot 2020-10-02 at 16 18 31

Is Person ID already linked?

nmcharlton:
Resolves #390 

Add planter registration date to planter detail dialog

Note: Most of the changes are whitespace (indent alignment)

Screenshot 2020-11-28 at 17 34 13

### DIFF
--- a/client/src/api/planters.js
+++ b/client/src/api/planters.js
@@ -6,14 +6,36 @@ import {
 import {session} from "../models/auth";
 
 export default {
-  getPlanter(id){
-    const query = `${process.env.REACT_APP_API_ROOT}/api/${getOrganization()}planter/${id}`;
-    return fetch(query, {
-      method: "GET",
-      headers: { 
-        "content-type": "application/json" ,
-        Authorization: session.token ,
-      },
+  getPlanter(id) {
+    const apiRoot = `${process.env.REACT_APP_API_ROOT}/api`;
+    const planterQuery = `${apiRoot}/${getOrganization()}planter/${id}`;
+    const registrationQuery = `${apiRoot}/planter_registration/?filter=where[planterId]=${id}`;
+
+    return Promise.all([
+      fetch(planterQuery, {
+        method: "GET",
+        headers: { 
+          "content-type": "application/json" ,
+          Authorization: session.token ,
+        },
+      }),
+      fetch(registrationQuery, {
+        method: "GET",
+        headers: { 
+          "content-type": "application/json" ,
+          Authorization: session.token ,
+        },
+      })
+    ])
+    .then(responses => {
+      let planter = responses[0]
+      const registrations = responses[1]
+      if (registrations.length) {
+        planter.lat = registrations[0].lat;
+        planter.lon = registrations[0].lon;
+        planter.createdAt = registrations[0].createdAt;
+      }
+      return planter;
     })
       .then(handleResponse)
       .catch(handleError);
--- a/client/src/api/planters.js
+++ b/client/src/api/planters.js
@@ -7,35 +7,14 @@ import {session} from "../models/auth";
 
 export default {
   getPlanter(id) {
-    const apiRoot = `${process.env.REACT_APP_API_ROOT}/api`;
-    const planterQuery = `${apiRoot}/${getOrganization()}planter/${id}`;
-    const registrationQuery = `${apiRoot}/planter_registration/?filter=where[planterId]=${id}`;
+    const planterQuery = `${process.env.REACT_APP_API_ROOT}/api/${getOrganization()}planter/${id}`;
 
-    return Promise.all([
-      fetch(planterQuery, {
-        method: "GET",
-        headers: { 
-          "content-type": "application/json" ,
-          Authorization: session.token ,
-        },
-      }),
-      fetch(registrationQuery, {
-        method: "GET",
-        headers: { 
-          "content-type": "application/json" ,
-          Authorization: session.token ,
-        },
-      })
-    ])
-    .then(responses => {
-      let planter = responses[0]
-      const registrations = responses[1]
-      if (registrations.length) {
-        planter.lat = registrations[0].lat;
-        planter.lon = registrations[0].lon;
-        planter.createdAt = registrations[0].createdAt;
-      }
-      return planter;
+    return fetch(planterQuery, {
+      method: "GET",
+      headers: { 
+        "content-type": "application/json" ,
+        Authorization: session.token ,
+      },
     })
       .then(handleResponse)
       .catch(handleError);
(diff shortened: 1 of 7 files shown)


server/src/models/planterRegistration.model.ts
```diff
@@ -0,0 +1,54 @@
...
+    type: 'number',
+  })
+  lat?: number;
+
+  @property({
+    type: 'number',
+  })
+  lon?: number;
+
+  // Define well-known properties here
```
mckornfield: is this comment relevant?

client/src/api/planters.js
```diff
@@ -80,4 +59,18 @@ export default {
...
     }).then(handleResponse).catch(handleError);
   },
+
+  getPlanterRegistrations(planterId) {
+    const registrationQuery =
+      `${process.env.REACT_APP_API_ROOT}/api/planter_registration?filter[where][planterId]=${planterId}`;
+    return fetch(registrationQuery, {
+      method: "GET",
+      headers: { 
+        "content-type": "application/json" ,
```
mckornfield, line 69: nit, remove quotes since Authorization doesn't have them, or add quotes to it

server/src/controllers/planterRegistration.controller.ts
```diff
@@ -35,7 +35,7 @@ export class PlanterRegistrationController {
     return await this.planterRepository.find(filter);
   }
 
-  @get('/planter/{id}', {
+  @get('/planter_registration/{id}', {
```
mckornfield: I don't know how many other urls we use, but I think it's preferable to use a dash instead of an underscore
For SEO as an example: https://developers.google.com/search/docs/advanced/guidelines/url-structure?hl=en&visit_id=637427852299866372-1076474063&rd=1

client/src/api/planters.js
```diff
@@ -6,14 +6,36 @@ import {
...
+      fetch(registrationQuery, {
+        method: "GET",
+        headers: { 
+          "content-type": "application/json" ,
+          Authorization: session.token ,
+        },
+      })
+    ])
+    .then(responses => {
+      let planter = responses[0]
```
mckornfield: I think this can be `const` right? you're just mutating it not reassigning

client/src/api/planters.js
```diff
@@ -6,14 +6,36 @@ import {
...
+  getPlanter(id) {
+    const apiRoot = `${process.env.REACT_APP_API_ROOT}/api`;
+    const planterQuery = `${apiRoot}/${getOrganization()}planter/${id}`;
+    const registrationQuery = `${apiRoot}/planter_registration/?filter=where[planterId]=${id}`;
+
+    return Promise.all([
+      fetch(planterQuery, {
+        method: "GET",
+        headers: { 
+          "content-type": "application/json" ,
```
mckornfield: same nitty- comments here and below about quotes

client/src/api/planters.js
```diff
@@ -6,14 +6,36 @@ import {
...
-    return fetch(query, {
-      method: "GET",
-      headers: { 
-        "content-type": "application/json" ,
-        Authorization: session.token ,
-      },
+  getPlanter(id) {
+    const apiRoot = `${process.env.REACT_APP_API_ROOT}/api`;
+    const planterQuery = `${apiRoot}/${getOrganization()}planter/${id}`;
+    const registrationQuery = `${apiRoot}/planter_registration/?filter=where[planterId]=${id}`;
```
mckornfield: same nit from the other place to use dashes in urls

server/src/controllers/planterRegistration.controller.ts
```diff
@@ -0,0 +1,48 @@
...
+import {PlanterRegistration} from '../models';
+import {PlanterRegistrationRepository} from '../repositories';
+
+export class PlanterRegistrationController {
+  constructor(
+    @repository(PlanterRegistrationRepository)
+    public planterRepository: PlanterRegistrationRepository,
+  ) {}
+
+  @get('/planter_registration', {
```
mckornfield: nit `planter-registration`

server/src/models/planterRegistration.model.ts
```diff
@@ -0,0 +1,54 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [prop: string]: any;
+
+  constructor(data?: Partial<PlanterRegistration>) {
+    super(data);
+  }
+}
+
+export interface PlanterRegistrationRelations {
+  // describe navigational properties here
```
mckornfield, line 55: is this for the future?

mckornfield (commented):
overall lg, might want to squash commits, will approve once we address the url (though maybe it doesn't matter, but I rarely see underscores in paths)

server/src/models/planterRegistration.model.ts
nmcharlton: Not really, just part of the template. I'll remove it.

client/src/api/planters.js
nmcharlton: I'll leave this one as it's done this way across the codebase – only using quotes in object keys where necessary (in this case, due to the hyphen).
It would benefit from a tidy up using `prettier`, though.

server/src/controllers/planterRegistration.controller.ts
nmcharlton: I agree. This was originally done for consistency with existing urls, but I'm inclined to move them all away from underscores.
I'll fix this one here and check with the team before changing the rest.

client/src/api/planters.js
mckornfield: ahh right, my bad, can ignore then

client/src/api/planters.js
nmcharlton: Yes, but it's been removed altogether in a subsequent commit.

server/src/controllers/planterRegistration.controller.ts
nmcharlton: Fixed.

server/src/models/planterRegistration.model.ts
nmcharlton: I suppose so. It's auto-generated by the Loopback CLI, but not yet used here.

client/src/api/planters.js
nmcharlton: Fixed

client/src/api/planters.js
nmcharlton: I'll ignore this one too 😄

mckornfield (approved):

PR #464, opened 2020-12-07
Show appropriate messages when users navigate to unauthorized or non-existent pages

### LINKED ISSUE #410: Logged in user should see 401 unauthorized access if route restricted, or 404
Logged in users should know that a route is restricted or unavailable.
Navigation attempts to a page should redirect to login, then direct back to the page after login.
Logged out users should always be directed to login.
Logged out users should not be able to probe to see which pages are 404 or restricted, but this is moot. Routes could be generated serverside...

nmcharlton:
Resolves #410 

Instead of redirecting to `/` when an incorrect route is entered, the user is given the appropriate 401 or 404 landing page.

Screenshot 2020-12-07 at 20 48 11
Screenshot 2020-12-07 at 20 48 36

### DIFF
--- a/client/src/components/Routers.js
+++ b/client/src/components/Routers.js
@@ -4,6 +4,8 @@ import Grid from '@material-ui/core/Grid'
 import Login from './Login'
 import { AppContext } from './Context'
 import PrivateRoute from './PrivateRoute'
+import Unauthorized from './Unauthorized'
+import Page404 from './Page404'
 
 export default function Routers() {
   const refContainer = React.useRef()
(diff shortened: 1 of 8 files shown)


client/src/components/Routers.js
```diff
@@ -31,23 +33,26 @@ export default function Routers() {
...
               }
               <Route
-                render={({location}) => (
-                  <Redirect to={{
-                    pathname: appContext.user ? "/" : "/login",
-                    state: { from: location }
-                  }}/>
-                )}
+                render={
+                  appContext.user ? (Page404) :
```
dadiorchen, line 48: Why haven't undefined user cause 404 page rather than unauthorized page?

nmcharlton: Undefined user (i.e. not logged in) should always redirect to `/login`.
The unauthorized page is reserved for logged in users trying to access something they don't have permission to see.

PR #482, opened 2020-12-14
Implement simple organization filter on Verify page

### LINKED ISSUE #442: Easy way to exclude organizations trees from main admin panel account
We as Greenstand need to be able to log in and by default hide the captures that come in from organisations that have their own login for verification. At the moment we (Greenstand) see the Freetown trees, OTP, Fairtree and other org trees which should only be visible by including them in the filter. That way we do not accidentally verify trees that are from other orgs. 

spec in progress from @deepwinter 
We are going to implement this as a default exclude filter based on entities (or just top level parent entities) that are in the entity relationship table. and will want to (eventually) have a UI to set this up. the default exclude filter should be per-user, persistant, and also be able to be disabled.

nmcharlton:
Resolves #442 

This is an initial attempt at an organization filter in Verify.

It lists the organizations (entries in the `entity` table with type `o`) in a single-selection dropdown include filter, along with _All_ and _Not set_.

The query reuses organization user tree filter logic and looks for trees with `planting_organization_id` or planted by planters with `organization_id` set to the selected organization.

Currently, other entity types are omitted from the selection list, and their trees don't appear when _Not set_ is selected.

Screenshot 2020-12-14 at 22 22 56

Screenshot 2020-12-14 at 22 23 03


dadiorchen (commented):
We are still not using the CI action for verify code? I saw the CI failed.

dadiorchen (approved):
I think this PR is good, though we are going far away with those tests I created :D https://github.com/Greenstand/treetracker-admin/blob/master/server/src/__tests__/integration/integration.ts
But writing tests on Loopback is awful!

sebastiangaertner:
This is also considering the entity_relationship ? Eg filter Freetown will include CBO1 CBO2 and CBO3

nmcharlton:
> This is also considering the entity_relationship ? Eg filter Freetown will include CBO1 CBO2 and CBO3

Yes, the trees associated with child organizations will currently be included in the filter for a parent organization.
@sebastiangaertner Do you consider this behaviour to be correct?

PR #496, opened 2021-02-03
feat: filter by not tagged on /verify

### LINKED ISSUE #489: Filter by "not tagged"
We need to be able to filter out the tagged trees (trees without species or custom tags). 
Use case: finding untagged trees in the system and give them at least one tag such as species or other tags. 
It will be two separate controls. (species and tags)

luacmartins:


client/src/components/FilterTop.js
```diff
@@ -10,9 +10,10 @@ import FilterModel, {
   SPECIES_NOT_SET,
   ALL_ORGANIZATIONS,
   ORGANIZATION_NOT_SET,
+  TAG_NOT_SET
 } from '../models/Filter';
 import DateFnsUtils from '@date-io/date-fns';
-import {connect} from 'react-redux';
+import { connect } from 'react-redux';
```
dadiorchen, line 16: I am realizing that previously, a contributor suggests that we can use prettier to force the code formatting, which might be correct, so we all are in a consistent way to code without the re-format annoy. :D

dadiorchen (approved):

client/src/components/FilterTop.js
luacmartins: I will try disabling vscode's auto formatting on save to minimize this issue.

dadiorchen: No, @luacmartins I don't think this is your problem at all, I think we should add a process to force format the code to solve this kind of problem, automatically.

nmcharlton: I agree. As long as it's applied consistently, this can only be a good thing.

PR #499, opened 2021-02-07
chore: fix broken EditPlanter test

nmcharlton:


dadiorchen (approved):
I'm also thinking about how to balance the effort costing and benefits gaining of fixing tests. Let's discuss this furtherly in the future. Especially considering that the test for component is more tedious than tests against like server-side.

PR #501, opened 2021-02-08
fix: audit logging

### LINKED ISSUE #491: Organization trees are not included in the Audit Log
Our audit log middleware intercepts API requests and logs the details of a request if its URL matches certain regexes.

https://github.com/Greenstand/treetracker-admin/blob/master/server/src/js/Audit.js#L93
```javascript
    } else if (/.*\/api\/trees\/\d+/.test(url)) {
```
This currently catches patches on `/api/trees/` but misses patches on trees associated with organizations (`/api/organization/{org_id}/trees/`).

It should catch all tree patches.

### LINKED ISSUE #492: Audit logging is not working
The process that writes logins and tree verifications to the `audit` table is failing for both actions.
- For logins, we're always returning code 401, even when login is successful.
- For tree verification actions, we are attempting to extract a user ID from the request that is not present.

luacmartins:
Should close #492

### DIFF
--- a/server/src/js/Audit.js
+++ b/server/src/js/Audit.js
@@ -26,6 +26,7 @@ export const auditMiddleware = (request, response, next) => {
         //console.log('req.header:', request.headers);
         //just audit when success
         //assert(response.statusCode);
+        console.log(request.url)
         if (/2\d\d/.test(response.statusCode)) {
           const audit = new Audit();
           await audit.did(request, response);
--- a/server/src/js/Audit.js
+++ b/server/src/js/Audit.js
@@ -7,7 +7,10 @@ import {Pool} from 'pg';
 // import log from 'loglevel';
 // import {strict as assert} from 'assert';
 import getDatasource from '../datasources/config';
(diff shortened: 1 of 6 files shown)


server/src/js/Audit.js
```diff
@@ -95,7 +96,8 @@ class Audit {
       //assert(req.method, req.method);
       //assert(req.user);
       //assert(req.user.id);
-      operator = req.user.id;
+      operator = JSON.parse(req.headers.user)
```
dadiorchen: I doubt this is the correct way to do so, @luacmartins your idea is to pass in the user info directly from client right? I think you don't need to do so, and this might also be unsafe. 
The idea is: parse the auth token from client, get the user info, inject into the req, and get it in the audit middleware, then put into the DB

dadiorchen (changes requested):

luacmartins:
Thanks for the input @dadiorchen! Hopefully this solves the issue.

nmcharlton (approved):
Looks good to me. Thanks @luacmartins!

PR #506, opened 2021-02-10
fix: 404 error at /check_session

### LINKED ISSUE #500: Resolve 404 error at the /check_session?id=1 endpoint
Error in browser console when accessed from local client

- createError.js:16 Uncaught (in promise) Error: Request failed with status code 404

gwynndp:
Remove check for dev environment from isAuth().
Update the auth to get back the user object so that the password hashs could be compared.
Make other minor fixes to typos and removed some console.logs.


nmcharlton (approved):
Thanks @gwynndp!

PR #510, opened 2021-02-13
fix: added logo and version placeholder to home page

jtaso:


jtaso:
@nmcharlton Do we want to pass the version number via an environment variable or some other mechanism?
Attaching a screenshot of an example of where the version and logo could be rendered.
Screenshot from 2021-02-14 06-10-48

nmcharlton:
@jtaso An environment variable is best.
We can add it to `client/.env` as described in this [Stack Overflow answer](https://stackoverflow.com/a/50822003/9929359), although the version for the app as a whole (server and client) is currently only set in the root `package.json`.
We will need to either replicate that version in `client/package.json` or use another approach.

nmcharlton:
@TANguyen1893 may be able to advise on the best way to make the version number available to the client; he recently implemented the auto-semantic versioning in the Admin Panel.

dadiorchen (approved):
THanks for bringing our logo back!

TANguyen1893:
With the current mono-repo setup, there is no clean way to propagate the version from the root package.json to the children package.json. Sure, we can add an environment variable in the .env file but there is no clean way to automatically populate it. Even when the app is deployed, it will not have access to the root package.json file so you won't be able to parse it at runtime.

I would prefer to split the client and server code to individual repos because the two projects have different release cycles.

ZavenArra:
> With the current mono-repo setup, there is no clean way to propagate the version from the root package.json to the children package.json. Sure, we can add an environment variable in the .env file but there is no clean way to automatically populate it. Even when the app is deployed, it will not have access to the root package.json file so you won't be able to parse it at runtime.
> 
> I would prefer to split the client and server code to individual repos because the two projects have different release cycles.

Admin panel team has indeed decided moving away from the mono-repo pattern is ideal, especially since there will soon be many other microservices that the admin panel will access.   @TANguyen1893 if you wanted to split the mono-repo now, you are welcome to do so, or we can wait until another dev gets to it.  The client side has already been configured to use dev-k8s as the default endpoint in preparation for this (so that frontend devs do not need to run the backend if they don't want to)

TANguyen1893:
Created this issue https://github.com/Greenstand/treetracker-admin/issues/520 for the separation of the client and server code. I assigned it to myself so I'll work on it.

TANguyen1893:
Any further development should be done in the admin panel client repo: https://github.com/Greenstand/treetracker-admin-client

jtaso:
Creating [new PR](https://github.com/Greenstand/treetracker-admin-client/pull/16) due to repo split.

PR #511, opened 2021-02-13
fix: allow users to list organizations

### LINKED ISSUE #509: API permissions are not configured for /organizations endpoint
The `/organizations` API endpoint does not have user permissions explicitly set in the server.
Querying this endpoint returns `401 Unauthorized`.

The symptom of this is that organizations are not listed in the _Verify_ filter or the _Edit Planter_ dialog.

This bug is masked in a development environment because we're skipping API permissions checks.

nmcharlton:
Resolves #509 

Users with `list_tree`, `approve_tree`, `manage_planter` or `super_user` permissions can access `/api/organizations` endpoint so that organizations can be listed in the _Verify_ and _Edit Planter_ tools.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -22,7 +22,7 @@
     "migrate": "node ./dist/migrate",
     "prestart": "npm run clean && npm run build",
     "start": "NODE_ENV=development node .",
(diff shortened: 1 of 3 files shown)


dadiorchen (approved):
Thanks for streamline these redundant and unreadable code written by me :D

PR #513, opened 2021-02-15
fix: timestamptz datatype spelling mistake

ghost:
- [x] replace `timestampz` with `timestamptz`

### DIFF
--- a/server/src/models/domainEvent.model.ts
+++ b/server/src/models/domainEvent.model.ts
@@ -45,7 +45,7 @@ export class DomainEvent extends Entity {
     required: true,
     postgresql: {
       columnName: 'created_at',
-      dataType: 'timestampz',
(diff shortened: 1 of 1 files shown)


nmcharlton (approved):

server/src/models/domainEvent.model.ts
```diff
@@ -45,7 +45,7 @@ export class DomainEvent extends Entity {
     required: true,
     postgresql: {
       columnName: 'created_at',
-      dataType: 'timestampz',
+      dataType: 'timestamptz',
```
dadiorchen, line 48: I want to learn from this: what's the difference between these two data type?

ghost: @dadiorchen it was giving me an unknown type error when migrating postgres data base.

PR #515, opened 2021-02-15
Enable verify publishing and adding rabbitmq url connection for deployment

arunbakt:
This change is to enable verify publishing by setting the corresponding environment variable to true and add rabbitmq connection url sealed secret for deployment in dev k8 cluster.


server/deployment/base/kustomization.yaml
```diff
@@ -4,3 +4,4 @@ resources:
 - treetracker-admin-api-service.yaml
```
dadiorchen, line 4: @arunbakt How do you deploy the service? I saw ppl add yaml in .github folder for deployment, how do you do it here?

dadiorchen (approved):

server/deployment/base/kustomization.yaml
arunbakt: @dadiorchen Yeah, the deployment folder seems to be used by the github workflow to deploy it to the k8 cluster.  I am still trying to understand the details on it.

PR #517, opened 2021-02-17
chore: move the sealedsecret file to the right location

arunbakt:
This is a minor update to move the rabbitmq sealedsecret file to the correct folder under /deployment.


arunbakt:
The change doesn't involve any code and addresses a deployment issue relating to sealed secret for rabbitmq. Hence taking the freedom to merge.

PR #519, opened 2021-02-17
fix: prettier config files for client/server

### LINKED ISSUE #508: Standardise use of prettier tool
We have the [prettier](https://prettier.io/) library configured in the project but its use is not automatically enforced and it is not set up consistently across `server` and `client` (in `*/.prettierrc`).

We need to do the following:

- Apply our style guide consistently across the project (as described in the README)
- Add a pre-commit hook to run `prettier src --write` in `client` and `server` to ensure the style is enforced

luacmartins:
Added instructions to enable Prettier on VSCode and made config files consistent on client and server. These changes might cause a lot of differences in new PRs, but hopefully this will add more formatting consistency to our code in the long run.

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -1,1 +1,6 @@
-{"hooks":{"commit-msg":"commitlint -E HUSKY_GIT_PARAMS"}}
+{
+  "hooks": {
+    "commit-msg":"commitlint -E HUSKY_GIT_PARAMS", 
+    "pre-commit": "pretty-quick --staged"
+    }
+}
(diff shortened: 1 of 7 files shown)


nmcharlton:
@luacmartins This all looks good. Can we add a pre-commit hook to automatically enforce the style when a file is committed? The process is described here: https://prettier.io/docs/en/precommit.html

We're already using husky to enforce the commit message format – pre-commit config can be added to `.huskyrc`.

dadiorchen (approved):

luacmartins:
@nmcharlton great suggestion! I believe it should be working with the pre-commit hook now.

nmcharlton (approved):
Awesome, thanks @luacmartins!

PR #523, opened 2021-02-18
feat: add image optimization to /verify and /planters

### LINKED ISSUE #454: optimizing performance through resizing images delivering them through CDN
Currently an image is displayed at around 480*420 px in a three column view before it jumps to the 4 column view. The actual image is delivered and loaded at around 768*1024px. The image could easily resized via CDN to about 1/3rd of its size in kB. Since most admin panel operators are in countries with slow internet this is a critical improvement and a low hanging fruit. Currently Freetown has to verify 50.000 trees and their internet is slow creating a bottle neck.

Possible services if not part of AWS or DO are https://statically.io/ 

Below two images ... one is whats loaded by the browser the other one is the size displayed. 

Screenshot 2020-11-20 at 06 10 54

luacmartins:
Sorry in advance for the formatting changes!

Adding image optimization to images on /verify, including plant details dialog and /planters, including planter details dialog. Closes #454.

The OptimizedImage component assumes that the original image urls follow the format below to parse the url and create a new one using Statically's CDN service. Please let me know if that is not the case for images in production.

https://{domain}.com/{image}

For example: 

https://treetracker-dev-images.s3.eu-central-1.amazonaws.com/2020.05.15.16.18.43_37.42138089_-121.87534965_27d083f2-e367-4ff3-b78b-06080b8b6b67_IMG_20200511_150539_5951525636141225822.jpg

### DIFF
--- a/client/src/components/OptimizedImage.js
+++ b/client/src/components/OptimizedImage.js
@@ -1,0 +1,56 @@
+import React from 'react';
+
+export default function OptimizedImage(props) {
+  const {
+    src,
+    width = 320,
+    height,
+    quality,
+    screenWidths = [1600, 1280, 960, 0],
+    imageSizes = [400, 300, 250, 200],
+    fixed,
+    ...rest
+  } = props;
+
+  if (!src) return <></>;
+
+  const cdnPath = 'https://cdn.statically.io/img';
+  const domain = src.match(/\/\/(.*)\//)[1];
+  const imagePath = src.match(/.com\/(.*)/)[1];
+  const params =
+    `f=auto,w=${width}` +
+    (height ? `,h=${height}` : '') +
+    (quality ? `,q=${quality}` : '');
+
+  const cdnUrl = `${cdnPath}/${domain}/${params}/${imagePath}`;
+
+  let sizes, srcSet;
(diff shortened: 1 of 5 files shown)


client/src/components/OptimizedImage.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
```
dadiorchen, line 1: So the image is generated on the fly /on demand? No pre-work? right? How about the performance when show the image lists.

luacmartins: Yes, images are generated on the fly and there is no upfront work required. We probably won't see much performance increase the first time an image is loaded, but the user will benefit from smaller file sizes. Each subsequent request should be much faster since the image will be cached. 

You can see the results below. Without optimization the total images size on the list page was 698kB and load time was 4.64s, with optimization total image size went down to 363kB and load time 3.60s.

Screen Shot 2021-02-18 at 9 59 21 AM
Screen Shot 2021-02-18 at 9 59 01 AM

luacmartins:
@nmcharlton Can I close this PR since it was merged on the client repo?

PR #529, opened 2021-02-21
fix: don't look for host ip from request headers while auditing

arunbakt:
Remove the code that looks for host in the request headers in audit.js. Corresponding test has been changed as well.

### DIFF
--- a/src/js/Audit.js
+++ b/src/js/Audit.js
@@ -3,7 +3,7 @@
 /*
  * To record operation on the system
  */
-import {Pool} from 'pg';
+import { Pool } from 'pg';
 // import log from 'loglevel';
 // import {strict as assert} from 'assert';
 import getDatasource from '../datasources/config';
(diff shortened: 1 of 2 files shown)


src/js/Audit.js
```diff
@@ -78,16 +78,15 @@ export const auditMiddleware = (request, response, next) => {
 
 class Audit {
   constructor() {
-    this.pool = new Pool({connectionString: getDatasource().url});
+    this.pool = new Pool({ connectionString: getDatasource().url });
```
arunbakt, line 81: these are prettier's changes I believe.

src/js/Audit.js
```diff
@@ -99,12 +98,14 @@ class Audit {
       //assert(res.myData);
       //assert(!isNaN(res.myData.user.id), res.myData.user.id);
       operator = res.myData.user.id; // change to operator = req.user.id
-    } else if (/(?:.*\/api\/trees\/\d+|.*\/api\/organization\/\d+\/trees\/\d+)/.test(url)) {
+    } else if (
```
arunbakt, line 101: this one is prettier's change as well I believe, please ignore

src/js/Audit.js
dadiorchen: :D you are a sevrious man.

dadiorchen (approved):

Issue #486, opened 2021-01-04
Can not delete species.


dadiorchen:
Reproduce: 

1. Open test.admin.treetracker.org 
2. Go to the species page
3. Find one species which's tagged count is zero.
4. Click 'delete', click 'confirm'
5. Doesn't work.

The request is complaining `500` error:

<img width="1552" alt="Screen Shot 2021-01-04 at 3 51 08 PM" src="https://user-images.githubusercontent.com/5744708/103513212-1becb480-4ea5-11eb-8009-e1e069ef7273.png">

nmcharlton:
Likely duplicate of #444

luacmartins:
This functionality seems to be working in my local environment, but not on test.admin.treetracker.org as mentioned. What are the differences between both environments?

nmcharlton:
> This functionality seems to be working in my local environment, but not on test.admin.treetracker.org as mentioned. What are the differences between both environments?

@luacmartins The DB policy in test and production prohibits `delete` instructions. Instead, we have an `active` column in all tables that is used to mark items as deleted. The front end should never show anything that is inactive.

See #444

nmcharlton:
This issue depends on Greenstand/treetracker-database#63

nmcharlton:
Fixed in #530

PR #530, opened 2021-02-25
feat: add /species/combine controller

luacmartins:
This PR is related to the combine species issue on the client side and creates the /species/combine endpoint.

### DIFF
--- a/src/controllers/species.controller.ts
+++ b/src/controllers/species.controller.ts
@@ -16,20 +16,22 @@ import {
   del,
   requestBody,
 } from '@loopback/rest';
-import {Species} from '../models';
-import {SpeciesRepository} from '../repositories';
+import { Species } from '../models';
+import { SpeciesRepository, TreesRepository } from '../repositories';
 
 export class SpeciesController {
   constructor(
     @repository(SpeciesRepository)
     public speciesRepository: SpeciesRepository,
+    @repository(TreesRepository)
+    public treesRepository: TreesRepository,
   ) {}
 
   @get('/species/count', {
     responses: {
       '200': {
         description: 'Species model count',
-        content: {'application/json': {schema: CountSchema}},
+        content: { 'application/json': { schema: CountSchema } },
       },
     },
   })
@@ -46,7 +48,7 @@ export class SpeciesController {
         description: 'Array of Species model instances',
         content: {
           'application/json': {
-            schema: {type: 'array', items: {'x-ts-type': Species}},
+            schema: { type: 'array', items: { 'x-ts-type': Species } },
           },
         },
       },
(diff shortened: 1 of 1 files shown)


src/controllers/species.controller.ts
```diff
@@ -86,6 +88,25 @@ export class SpeciesController {
...
+        description: 'Species POST success',
+      },
+    },
+  })
+  async combine(
+    @requestBody() request: { combine: number[]; species: Species },
+  ): Promise<void> {
+    const newSpecies = await this.speciesRepository.create(request.species);
+    console.log(newSpecies);
+    for (const id of request.combine) {
```
dadiorchen: Seems with Loopback, is easy to wrap the executions in a transaction, for this `for` loop, it might bring some safety. 
Or, could we think about use bulk update to reduce the connection count to the DB? Like: `update set a=x where id in (1,2,3)`

dadiorchen (approved):
Another question, how do we verify it works? Make up some data in DB and request the endpoint manually?

src/controllers/species.controller.ts
luacmartins: That sounds interesting! I will try to implement the bulk update version. Thanks for the feedback.

For the testing, I was using it in parallel with the client to manually add/combine species and also using Postman. Do we have any other way to test it?

src/controllers/species.controller.ts
```diff
@@ -86,6 +88,25 @@ export class SpeciesController {
...
+      const query = `UPDATE trees SET species_id=${newSpecies.id} WHERE species_id=${id}`;
+      await this.treesRepository.execute(query, []);
+      await this.speciesRepository.deleteById(id);
```
nmcharlton: Unfortunately, this won't work in test and production environments, due to deletes being prohibited (issue #486). We should really set `active=false`.

nmcharlton (changes requested):
It might be worth fixing #486 at the same time to ensure consistent delete behaviour.

src/controllers/species.controller.ts
luacmartins: Makes sense. Although active is an integer column. Are we using 0 for inactive and 1 for active?

nmcharlton: I think that must have been implemented incorrectly - it should be a Boolean. We'll need to raise a bug against the database project ([treetracker-database](https://github.com/Greenstand/treetracker-database)) and create a database migration script to correct it.

luacmartins: I talked to @deepwinter and filed an issue there. I'm sending another commit with the updated code but we should wait until changes are made to the database to merge it.

nmcharlton (approved):
Lovely work 👍

PR #533, opened 2021-03-03
chore: create api deploy workflows

nmcharlton:
Based on webmap deployment workflows


dadiorchen (approved):
Ok, something I don't know

PR #537, opened 2021-03-11
Feat: Add country

luacmartins:
Changes the /registrations route to add the country of registration of each planter.

### DIFF
--- a/src/controllers/planterRegistration.controller.ts
+++ b/src/controllers/planterRegistration.controller.ts
@@ -1,14 +1,7 @@
-import {
-  Filter,
-  repository,
-} from '@loopback/repository';
-import {
-  param,
-  get,
-  getFilterSchemaFor,
-} from '@loopback/rest';
-import {PlanterRegistration} from '../models';
-import {PlanterRegistrationRepository} from '../repositories';
+import { Filter, repository } from '@loopback/repository';
+import { param, get, getFilterSchemaFor } from '@loopback/rest';
+import { PlanterRegistration } from '../models';
+import { PlanterRegistrationRepository } from '../repositories';
 
 export class PlanterRegistrationController {
   constructor(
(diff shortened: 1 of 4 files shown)


src/controllers/planterRegistration.controller.ts
```diff
@@ -1,14 +1,7 @@
...
-  repository,
-} from '@loopback/repository';
-import {
-  param,
-  get,
-  getFilterSchemaFor,
-} from '@loopback/rest';
-import {PlanterRegistration} from '../models';
-import {PlanterRegistrationRepository} from '../repositories';
+import { Filter, repository } from '@loopback/repository';
```
nmcharlton: We can remove the unused imports.

src/controllers/planterRegistration.controller.ts
```diff
@@ -22,28 +15,44 @@ export class PlanterRegistrationController {
...
         },
       },
     },
   })
   async find(
-    @param.query.object('filter', getFilterSchemaFor(PlanterRegistration))
-    filter?: Filter<PlanterRegistration>,
+    @param.query.object('filter') filter: object,
   ): Promise<PlanterRegistration[]> {
-    return await this.planterRepository.find(filter);
```
nmcharlton, line 35: I'm a little concerned about removing the full support for `filter`.
If someone queries this endpoint with parameters other than `where.planterId`, they will be silently ignored.

There's a slightly convoluted example in [`trees.controller.js`](https://github.com/Greenstand/treetracker-admin-api/blob/46583de2ff47d3d0914e9de6a53a7e26894cc3ef/src/controllers/trees.controller.ts#L273) of how we can augment a loopback query.

You'd just need to do the `where` bits to build the `WHERE` clause.

nmcharlton (changes requested):
Looks good, just a suggestion on the planter registration filter.

dadiorchen (commented):
Great job, @luacmartins

src/controllers/planterRegistration.controller.ts
luacmartins: I tried to add support for `filter` as suggested. Please let me know if this is the right way to do it.
I also abstracted `buildFilterQuery` to its own file and replaced it in `trees.controller.ts` so we don't repeat ourselves.

Thanks for the feedback!

src/controllers/planterRegistration.controller.ts
```diff
@@ -22,28 +25,51 @@ export class PlanterRegistrationController {
...
+              items: { 'x-ts-type': PlanterRegistration },
+            },
           },
         },
       },
     },
   })
   async find(
     @param.query.object('filter', getFilterSchemaFor(PlanterRegistration))
-    filter?: Filter<PlanterRegistration>,
```
nmcharlton, line 33: `Filter<PlanterRegistration>` will do the job here.

src/controllers/planterRegistration.controller.ts
```diff
@@ -1,14 +1,17 @@
...
-import {PlanterRegistrationRepository} from '../repositories';
+import { Filter, repository, Where } from '@loopback/repository';
+import { param, get, getFilterSchemaFor } from '@loopback/rest';
+import { PlanterRegistration } from '../models';
+import { PlanterRegistrationRepository } from '../repositories';
+import { buildFilterQuery } from '../js/buildFilterQuery.js';
+
+// Extend the LoopBack filter types for the Trees model to include tagId
+// This is a workaround for the lack of proper join support in LoopBack
+type PlanterRegistrationWhere = Where<PlanterRegistration> & {
```
nmcharlton: There's no need to extend the `Where` and `Filter` objects here – `planterId` is part of the `PlanterRegistration` model already.

src/controllers/planterRegistration.controller.ts
nmcharlton: Awesome! Nice work.

nmcharlton (commented):
Just one minor comment this time – a bit of redundant code.

src/controllers/planterRegistration.controller.ts
luacmartins: Thanks for all the great feedback @nmcharlton! I really appreciate it.

nmcharlton (approved):
Excellent work as always. Thanks @luacmartins!

PR #541, opened 2021-03-14
fix: exclude inactive species

### LINKED ISSUE #540: Inactive species are included in API responses
Species with `active = false` should be treated by the API as deleted and not included in `find` and `count` queries.

nmcharlton:
Resolves #540


dadiorchen (commented):
✌️

PR #543, opened 2021-03-27
fix: user roles and policies

### LINKED ISSUE #445: Only first role assigned to a user is queried for policies
Users can be assigned multiple roles (Admin, Tree Manager etc.) and each role carries a set of permission policies that determine what the user can see and do on the Admin Panel.

We currently only query policies from the first role assigned to the user.
We should combine policies from all of their roles.

### LINKED ISSUE #536: Non-admin users do not have their roles listed on Account page
If a user does not have the admin role, querying their own roles on the Account page fails with a 401 Unauthorized.

Screenshot 2021-01-09 at 16 23 34

### LINKED ISSUE #544: Cannot update user roles
Adding roles to a user in the User Manager tool is broken.
The API throws an exception:
```
ReferenceError: addAdminUserRole is not defined
```

This is the offending line of code in `src/js/auth.js`:
```js
await addAdminUserRole(req.params.userId, req.body.role[i]);
```

This function was moved to a helper object for testing but this instance was somehow missed.

### LINKED ISSUE #545: Organization user permissions are broken
When an organization user logs in, the API returns 401 Access Denied for all trees/planter API calls.
This is due to a rogue `hasPermission()` function in `src/js/auth.js`:
```js
const hasPermission = (userPolicies, allowedPolicies, orgReq, organization) => {
  if (orgReq && organization && organization.id > 0) {
    return false;
  }

  return userPolicies.some((r) => allowedPolicies.some((p) => r.name === p));
};
```

This is clearly incorrect – we always return `false` if the user has an organization.
The correct behaviour is to check that the organization ID in the request matches that of the user.

nmcharlton:
This PR includes four bug fixes:
- Resolves #445 
- Resolves #536
- Resolves #544 
- Resolves #545 

## Listing User Roles for Non-Admins
The account page now shows all users all of the roles assigned to them.

## Users with Multiple Roles
Policies from _all_ roles assigned to a user are now combined to determine access permissions.
A workaround has been added to prevent users being assigned roles associated with multiple organizations – the admin UI allows roles associated with a maximum of one org role to be selected.

Screenshot 2021-03-27 at 19 47 29

## Updating User Roles
Existing users can now have their roles updated.

## Organization Permissions
Organization permissions are now correctly applied. Org users can access trees and planters from their org, but no other org. Non-org users can access any trees and planters.

### DIFF
--- a/.env.development.orig
+++ b/.env.development.orig
@@ -1,0 +1,1 @@
+DATABASE_URL="postgresql://doadmin:l5al4hwte8qmj6x8@db-postgresql-sfo2-nextgen-do-user-1067699-0.db.ondigitalocean.com:25060/treetracker_dev?ssl=true"
--- a/.env.development.orig
+++ b/.env.development.orig
@@ -1,1 +1,0 @@
-DATABASE_URL="postgresql://doadmin:l5al4hwte8qmj6x8@db-postgresql-sfo2-nextgen-do-user-1067699-0.db.ondigitalocean.com:25060/treetracker_dev?ssl=true"
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "migrate": "node ./dist/migrate",
     "prestart": "npm run clean && npm run build",
     "start": "NODE_ENV=development node .",
-    "start:debug": "NODE_ENV=development node --inspect .",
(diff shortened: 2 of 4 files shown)


nmcharlton:
Most of the changed lines are whitespace changes by Prettier.

dadiorchen (approved):

PR #557, opened 2021-04-29
fix: fixed setting organizationID to null value

### LINKED ISSUE #556: Organization ID cannot be set to null in Planter
Greenstand/treetracker-admin-client#71 aims to allow the user to set a planter's organization ID to `null`, but the current `planter` LoopBack model does not fully allow `null`:

```
  @property({
    type: Number,
    required: false,
    scale: 0,
    postgresql: {
      columnName: 'organization_id',
      dataType: 'integer',
      dataLength: null,
      dataPrecision: null,
      dataScale: 0,
      nullable: 'YES',
    },
  })
  organizationId?: Number;
```

According to https://github.com/strongloop/loopback-next/issues/2220#issuecomment-524567458, adding the following to the `@property` decorator fixes the problem:
```
jsonSchema: {
  nullable: true,
},
```

tranquanghuy0801:
Resolves #556 

Added `jsonSchema: {
      nullable: true,
    },` to the @property of **organizationId**

### DIFF
--- a/src/models/planter.model.ts
+++ b/src/models/planter.model.ts
@@ -152,6 +152,9 @@ export class Planter extends Entity {
   @property({
     type: Number,
     required: false,
+    jsonSchema: {
+      nullable: true,
+    },
     scale: 0,
     postgresql: {
       columnName: 'organization_id',


nmcharlton (approved):
Perfect

PR #562, opened 2021-05-16
fix: organization user unauthorized error for verify filter

gwynndp:
auth.js
        - add a new path to check for organization permissions
        - adjust logic to allow both admin and org accounts to get permissions based on if a requestedOrgId is in the url
        
        https://github.com/Greenstand/treetracker-admin-client/issues/70

### DIFF
--- a/src/controllers/planterOrganization.controller.ts
+++ b/src/controllers/planterOrganization.controller.ts
@@ -128,5 +128,13 @@ export class PlanterOrganizationController {
         'Organizational user has no permission to do this operation',
       );
     }
+    // const result = await this.planterRepository.findById(id);
+    // const entityIds = await this.treesRepository.getEntityIdsByOrganizationId(organizationId);
+    // if (!entityIds.includes(result?.organizationId?.valueOf() || -1)) {
+    //   throw new HttpErrors.Unauthorized(
+    //     'Organizational user has no permission to do this operation',
+    //   );
+    // }
+    // await this.planterRepository.updateById(id, planter);
   }
 }
--- a/src/controllers/planterOrganization.controller.ts
+++ b/src/controllers/planterOrganization.controller.ts
@@ -108,7 +108,7 @@ export class PlanterOrganizationController {
         'Organizational user has no permission to do this operation',
       );
     }
-    return await this.planterRepository.findById(id);
+    return result;
   }
 
   @patch('/organization/{organizationId}/planter/{id}', {
@@ -123,18 +123,10 @@ export class PlanterOrganizationController {
     @param.path.number('id') id: Number,
     @requestBody() planter: Planter,
   ): Promise<void> {
-    throw new HttpErrors.Unauthorized(
-      'Organizational user has no permission to do this operation',
-    );
-    const result = await this.planterRepository.findById(id);
-    const entityIds = await this.treesRepository.getEntityIdsByOrganizationId(
-      organizationId.valueOf(),
-    );
-    if (!entityIds.includes(result?.organizationId?.valueOf() || -1)) {
+    if (organizationId) {
       throw new HttpErrors.Unauthorized(
         'Organizational user has no permission to do this operation',
       );
     }
-    await this.planterRepository.updateById(id, planter);
   }
 }
--- a/src/js/auth.js
+++ b/src/js/auth.js
@@ -161,11 +161,7 @@ helper.hasPermission = (
   allowedPolicyNames,
   requestedOrgId,
 ) => {
-  if (
-    userOrganization &&
-    requestedOrgId &&
-    requestedOrgId !== userOrganization.id
-  ) {
+  if (userOrganization && requestedOrgId !== userOrganization.id) {
     return false;
   }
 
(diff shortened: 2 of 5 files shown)


src/js/auth.js
```diff
@@ -545,6 +543,30 @@ const isAuth = async (req, res, next) => {
         return next();
       }
 
+      matcher = url.match(/\/api\/organizations\.*/);
```
nmcharlton: I'm not sure what this is doing that isn't already handled by the organization user url matcher on line 618 below.
If we're going to allow an organization user to query organizations, they will need to be restricted to querying/updating their own org and child orgs.

src/controllers/planterOrganization.controller.ts
```diff
@@ -123,18 +123,10 @@ export class PlanterOrganizationController {
...
   ): Promise<void> {
-    throw new HttpErrors.Unauthorized(
-      'Organizational user has no permission to do this operation',
-    );
-    const result = await this.planterRepository.findById(id);
-    const entityIds = await this.treesRepository.getEntityIdsByOrganizationId(
-      organizationId.valueOf(),
-    );
-    if (!entityIds.includes(result?.organizationId?.valueOf() || -1)) {
+    if (organizationId) {
```
nmcharlton, line 126: We need to fix this properly – either org users should be able to update their own planters' details, or the edit functionality should be disabled for org users in the client and only available to top-level planter managers.

src/js/auth.js
```diff
@@ -161,7 +161,11 @@ helper.hasPermission = (
   allowedPolicyNames,
   requestedOrgId,
 ) => {
-  if (userOrganization && requestedOrgId !== userOrganization.id) {
+  if (
+    userOrganization &&
+    requestedOrgId &&
```
nmcharlton: A valid `userOrganization` and a falsy `requestedOrgId` means that the user is trying to request data for _all_ orgs, not just their own.
This operation should be blocked.

src/js/auth.test.js
```diff
@@ -112,7 +112,7 @@ describe('auth', () => {
       );
       expect(hasPermission([{ name: 'test' }], null, ['test'], 1)).toBe(true);
       expect(hasPermission([{ name: 'test' }], { id: 1 }, ['test'], null)).toBe(
-        false,
+        true,
```
nmcharlton: An org user should not have permission to query a non-org endpoint, which is what this function is checking.

nmcharlton (changes requested):
Unfortunately, these changes remove some of our org user security measures.
At some point, we need to allow org users to query sub-org (in much the same way that they can query trees and planters assigned to their org), but allowing them to query _all_ orgs like this is a potential security risk.

src/js/auth.js
gwynndp: The requestedOrgId doesn't exist for the admin role.  Should we refactor admin login so we can require a requestedOrgId in all cases?

src/js/auth.js
gwynndp: Adding this matcher was a compromise because the client wasn't passing along the org id in its request ("organizations?filter[where][type]=O") and the original matcher pattern was looking for it, but neither the organization.controller nor the planterOrganization.controller have an endpoint for that pattern, "organization/organizationId/organizations".

So the original matcher works for permissions but the request returns a 404 because the endpoint doesn't exist.

Alternatives: 
We can add an endpoint for "organization/organizationId/organizations" to one of the controllers or pass along the org Id in a filter to the organization.controller

If we want to require all logins to have a orgId then we might update the organization.controller endpoints

src/js/auth.js
nmcharlton: Currently, all org requests for trees/planters follow the `/api/organization/{requestedOrgId}/trees|planter/` pattern, while all non-org requests use `/api/trees|planter/` (with no `requestedOrgId`).

The intention is that non-org users could make requests for specific orgs using the `/api/organization/{requestedOrgId}/...` pattern (in fact, this might be the better way of doing filtering by org in the Verify, Planters and Trees pages).

The best fix would be to implement the same organization-specific pattern for organization queries (possibly with a name change to reduce confusion).
There will soon be a new `stakeholder` endpoint in the Treetracker API (or possibly a separate service) that will allow us to query and manage organizations / entities / stakeholders, so I'm reluctant to do too much more on this in the legacy API.

What do you think?

src/js/auth.test.js
gwynndp: Makes sense.   I'd reasoned only admin would have a null value for requestedOrgId.

src/js/auth.js
gwynndp: I misunderstood the intent as allowing admin to have access when the requestedOrgId was not provided.

src/controllers/planterOrganization.controller.ts
gwynndp: New issue once decision is made about direction?

src/js/auth.js
gwynndp: I tried out adding a new endpoint for organizations that would filter by entity.  I also updated the request from the client to pass along the orgId from the get-go.  It seems to be working so far.  I'll push the client changes in a little bit.

src/js/auth.test.js
nmcharlton: Unfortunately, that part of the project is not very intuitive.
I wonder whether it would be cleaner to require every user to be assigned an org ID, with Greenstand as a top-level org and all others as child orgs.

nmcharlton (approved):
Nice work 👌

PR #564, opened 2021-05-21
Enable verify publish to connect to rabbitmq and publish messages in dev

arunbakt:
The change uses kustomize overlay to patch the env variable for enabling message publishing on verifications.


arunbakt:
I don't know how to withdraw this pull request, so closing it.

PR #565, opened 2021-05-21
Enable verify publishing env variable

arunbakt:
Restore ENABLE_VERIFY_PUBLISHING env variable to the base deployment file and set it to false.
Using overlays, set this value to true in development environment that will trigger each verification on admin panel to publish messages for field-data app to consume.


nmcharlton (approved):

nmcharlton:
Thanks @arunbakt!

PR #568, opened 2021-05-21
chore: move the rabbitmq connection inside try/catch block

arunbakt:
This is a quick fix to capture any error that might be occuring when connecting to rabbitmq service by moving the connection within a try/catch block.

I am taking the liberty to merge this code to test this out to see if I get any errors when publishing messages.


arunbakt:
I am merging this PR without seeking a review (the code change is safe and you could take a look at the change) to enable me to troubleshoot unpublished messages and surface any errors that might be happening in the dev systems.

PR #569, opened 2021-05-22
chore: Use equivalence operator for the verify publishing flag

arunbakt:
The current check the verify publishing flag always succeeds and results in message publishing to happen even if the value is set to false. This change fixes that to only succeed if the value of the flag is equal to the string value 'true'.

Note: Currently tree verifications in dev instance are failing due to an error ( i suspect it is to do with connection issue with rabbit mq). This failure shouldn't cause problem in prod since the verify publishing flag is set to false. I have locally tested it with the flag set to true and false with an invalid rabbitmq connection parameter. The value set to false doesn't fail any operation in the admin panel. So this change is crucial so that prod instances don't get affected if this code makes it to prod. I will work with @ZavenArra to see why the dev instance is throwing the error I see when performing  tree verifications.


src/controllers/trees.controller.ts
```diff
@@ -171,7 +171,7 @@ export class TreesController {
     try {
       let verifyCaptureProcessed;
       let domainEvent;
-      if (config.enableVerificationPublishing) {
+      if (config.enableVerificationPublishing === 'true') {
```
nmcharlton: I'd suggest using `config.enableVerificationPublishing?.toLowerCase()` here, to accommodate people setting the env var to `TRUE` or `True`.

An even better option would be to implement a simple, reusable env var parser in `config.js` to make `enableVerificationPublishing` a boolean and avoid the need for a string literal here.

nmcharlton (changes requested):

src/controllers/trees.controller.ts
arunbakt: @nmcharlton , Updated the PR with your suggestions. Please take a look when you get a chance!

nmcharlton (approved):

PR #570, opened 2021-05-23
fix: add org id to planter endpoints to so orgs can edit planters

gwynndp:


src/controllers/planter.controller.ts
```diff
@@ -74,14 +74,15 @@ export class PlanterController {
     return await this.planterRepository.findById(id);
   }
 
-  @get('/planter/{id}/selfies', {
+  @get('/organization/{organizationId}/planter/{id}/selfies', {
```
nmcharlton: This endpoint is already implemented in `planterOrganization.controller.ts`.

src/controllers/planterRegistration.controller.ts
```diff
@@ -10,7 +10,7 @@ export class PlanterRegistrationController {
     public planterRepository: PlanterRegistrationRepository,
   ) {}
 
-  @get('/planter-registration', {
+  @get('/organization/{organizationId}/planter-registration', {
```
nmcharlton: I think we'll need to keep the original endpoint for non-org requests.
The `trees` and `planter` org endpoints are implemented in separate controllers: `treesOrganization.controller.ts` and `planterOrganization.controller.ts` respectively.

src/controllers/planter.controller.ts
gwynndp: The existing endpoint wasn't responding to the "/selfies" request.  I can add the /selfies endpoint with the org id to the planterOrganization.controller.ts and keep the non-org endpoint on planter.controller.

src/controllers/planterRegistration.controller.ts
gwynndp: Agreed. I can add the /planter-registration endpoint with the org id to the planterOrganization.controller.ts and keep the non-org endpoint on planter.controller.

nmcharlton (approved):
Great work, as always 🚀

PR #572, opened 2021-05-31
feat: add token_id to trees model

### LINKED ISSUE #571: Add tokenId to tree record
In order to support Greenstand/treetracker-admin-client#53, the `tokenId` from the `trees` record needs to be included in the `trees` Loopback model.

nmcharlton:
Resolves #571

### DIFF
--- a/src/models/trees.model.ts
+++ b/src/models/trees.model.ts
@@ -364,14 +364,24 @@ export class Trees extends Entity {
   plantingOrganizationId?: Number;
 
   @property({
-    type: Number,
+    type: String,
     required: false,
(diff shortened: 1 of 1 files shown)


dadiorchen (commented):
I don't know too much about the Postgresql and the data type spec, I found zaven is using uuid for the token_id, just let you know @nickharriscodes for a future head-up in advance.

https://github.com/Greenstand/treetracker-wallet-api/blob/fe71d8e74cde60605f62c9a44209b165c93b7eea/database/migrations/20200901213253-CreateTableToken.js#L19-L22

PR #577, opened 2021-06-16
fix: include capture count in species results

### LINKED ISSUE #573: Per-species capture count queries are very slow
When the Admin Panel loads, the species model queries the capture count for each species.
In production at the time of writing, there are around 40 species and 1,000,000 captures (these numbers will only rise).
The 40 parallel requests take around 10 seconds to complete.

nmcharlton:
Resolves #573

This PR merges the species capture counts directly into the results of the standard species find query.

Unfortunately, LoopBack makes it quite fiddly to arbitrarily extend objects in API responses, so I've returned `Promise<any[]>` for simplicity.
I'm open to suggestions for a better solution.

Ideally, we'd do the species find and get the capture counts all in one SQL query, but LoopBack makes the former a little opaque.
That said, performing two separate queries and merging manually shouldn't add significant overhead.

I'll update the corresponding query in the client to not make separate `api/trees/count` queries (under Greenstand/treetracker-admin-client#115).

### DIFF
--- a/src/controllers/species.controller.ts
+++ b/src/controllers/species.controller.ts
@@ -66,7 +66,8 @@ export class SpeciesController {
     filter = { ...filter, where: { ...filter?.where, active: true } };
 
     const captureCountStmt =
-      'SELECT species_id,COUNT(species_id) as count FROM trees WHERE species_id NOTNULL GROUP BY species_id';
+      `SELECT species_id,COUNT(species_id) as count FROM trees ` +
+      `WHERE species_id NOTNULL AND active=true GROUP BY species_id`;
     // TODO: combine both queries in one query to minimise overhead
     const [speciesList, captureCounts] = await Promise.all([
       this.speciesRepository.find(filter),
--- a/src/controllers/species.controller.ts
+++ b/src/controllers/species.controller.ts
@@ -58,12 +58,31 @@ export class SpeciesController {
   })
   async find(
     @param.query.object('filter', getFilterSchemaFor(Species))
-    filter?: Filter<Species>,
-  ): Promise<Species[]> {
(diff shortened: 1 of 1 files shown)


src/controllers/species.controller.ts
```diff
@@ -58,12 +58,32 @@ export class SpeciesController {
   })
   async find(
     @param.query.object('filter', getFilterSchemaFor(Species))
-    filter?: Filter<Species>,
-  ): Promise<Species[]> {
+    filter?: Filter<Species> & { fields?: { captureCount?: Boolean } },
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ): Promise<any[]> {
```
dadiorchen, line 63: In my opinion, it's fine not struggling with the accurate type.

src/controllers/species.controller.ts
```diff
@@ -58,12 +58,32 @@ export class SpeciesController {
...
     console.log('get /species filter --> ', filter ? filter.where : null);
     // Only include active species
     filter = { ...filter, where: { ...filter?.where, active: true } };
-    return await this.speciesRepository.find(filter);
+
+    const captureCountStmt =
+      `SELECT species_id,COUNT(species_id) as count FROM trees ` +
+      `WHERE species_id NOTNULL AND active=true GROUP BY species_id`;
```
dadiorchen, line 70: How long time this sql would take?

dadiorchen: Roughly

dadiorchen (approved):

src/controllers/species.controller.ts
nmcharlton: In the dev DB it's roughly 200ms in total

PR #582, opened 2021-07-18
docs: create OpenAPI spec

### LINKED ISSUE #581: Generate OpenAPI spec
The Admin Panel API needs to be documented in the same way as other Greenstand APIs using OpenAPI 3.0. 

This spec can then be linked to from the Admin Panel client guidance.

nmcharlton:
Resolves #581 

OpenAPI 3.0.0 spec for Admin API generated using the `openapi.json` file served by LoopBack as part of its API explorer.
This file was then converted to yaml using SwaggerHub's import feature.

Note that it only covers the half of the API served by LoopBack, but I want to get this in as a starting point.
The `/auth` side is not yet included – I'll raise a separate issue to cover that one.

### DIFF
--- a/docs/api/spec/treetracker-admin.v1.yaml
+++ b/docs/api/spec/treetracker-admin.v1.yaml
@@ -1,0 +1,2135 @@
+openapi: 3.0.0
+info:
+  title: Treetracker Admin API
+  version: 2.17.4
+paths:
+  '/organization/{organizationId}/organizations':
+    get:
+      x-controller-name: OrganizationController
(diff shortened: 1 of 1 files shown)


dadiorchen:
So Nick are we going to offer this to 3thd party org outside Greestand?

nmcharlton:
> So Nick are we going to offer this to 3thd party org outside Greestand? 

There are no plans to do so at the moment. This is just intended to support Admin Panel client developers.

dadiorchen (approved):

PR #589, opened 2021-09-20
fix: planter filter to allow filtering parent orgs by name and all orgs by id

gwynndp:
- updated /planter and /count endpoints to get both parent and child org entityIds and and planter ids for the query
- updated /organization/{orgid}/planter, .../trees, and .../count endpoints to check if an org filter was requested, compare it to the org account id, and override the value in the query if it's a sub-org
- updated /organizations to get all the parent orgs for admin and all the sub-orgs for an org account

### DIFF
--- a/src/controllers/organization.controller.ts
+++ b/src/controllers/organization.controller.ts
@@ -14,6 +14,11 @@ import {
 import { Organization } from '../models';
 import { OrganizationRepository } from '../repositories';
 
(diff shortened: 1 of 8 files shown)


dadiorchen (approved):

nmcharlton:
My only concern is that there appears to be a lot of duplication.

PR #594, opened 2021-10-01
fix: convert planter attributes to camel case

### LINKED ISSUE #593: Planter attributes are snake case instead of camel case
The convention with LoopBack is to use snake case for database column names and convert them to camel case for model attributes.
If we're executing scripts directly rather than using LoopBack wrappers (such as `repository.find()`), we need to explicitly convert the column names to the corresponding model attributes as expected by the client.

The result of snake case is that several planter attributes are missing in the UI, including photo and name.
Screenshot 2021-10-01 at 21 59 19

nmcharlton:
Resolves #593 

Planter attributes are now returned as expected by the client:
Screenshot 2021-10-01 at 22 19 00

Also tested with an organization user:
Screenshot 2021-10-01 at 22 20 43


gwynndp (approved):
Looking good!

PR #595, opened 2021-10-08
Node16 update + Bug fix for Growers org-queries

### LINKED ISSUE #591: Upgrade to Node 16
The upgrade to Node 16 is now critical, as we cannot release new versions without it.
The project should be built and tested against Node 16.

gwynndp:
Node 16 updates:
- updated loopback packages, typescript, and ts-jest
- added pg module to help manage ssl connection to Postgres database (this will require leaving the "ssl=true" off of the end of the database url from now on as that setting option was deprecated)
- update the sequence.ts code for updated loopback

Bug Fix:
- fix query building for Growers so that it accesses the correct fields in organization related queries

Closes: https://github.com/Greenstand/treetracker-admin-api/issues/591


package.json
```diff
@@ -97,8 +98,8 @@
     "prettier": "^2.2.1",
     "pretty-quick": "^3.1.0",
     "supertest": "^4.0.2",
-    "ts-jest": "^26.1.4",
-    "typescript": "^3.9.7"
+    "ts-jest": "^26.5.6",
+    "typescript": "~4.3.5"
```
dadiorchen, line 102: @gwynndp we have to upgrade all of these libs if we want to use node.js 16?

dadiorchen (approved):

package.json
gwynndp: @dadiorchen The Admin Panel API uses Loopback and that needed to be updated for node 16.  Typescript and ts-jest were required to update because of dependencies in Loopback.  Prettier probably didn't need the update, but it doesn't hurt either way.

PR #597, opened 2021-10-27
refactor: create mixin to share helper functions across repos

gwynndp:
## Description
Organization of helper functions to reduce duplication across repositories

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines

**What kind of change does this PR introduce?** 
 - [x] Enhancement

## Issue

**What is the current behavior?**

There are duplicate functions to help build queries for filtering by organization information in the Trees, Planter, and Organization repositories

**What is the new behavior?**

Can now use a mixin with a repository to import the helper functions from utils.repository-mixin.ts

Steps taken:
 - [x]  Created utils.repository-mixin.ts to house helper functions to integrate organization info into filter queries for use by repositories
 - [x]  Moved helper functions from Trees, Planter, & Organization repositories to utils.repository-mixin.ts to reduce duplication
 - [x]  added parameters to handle the Trees and Planters query builds differently based on different table names for the organization and id fields
 - [x]  update the controllers to pass the necessary information for building the queries
 - [x]  fix the query for when the organization has a value of null, represented as "Not Set" in the UI dropdown

## Breaking change
**Does this PR introduce a breaking change?** 
 - [x] No

### DIFF
--- a/src/js/buildFilterQuery.js
+++ b/src/js/buildFilterQuery.js
@@ -28,6 +28,20 @@ export function buildFilterQuery(selectStmt, params) {
 
         const whereObjClause = connector._buildWhere(modelName, safeWhere);
 
+        //add the modelName to each requested field to avoid ambiguity in more complex queries
+        console.log('whereObjClause.sql -------> ', whereObjClause.sql);
+
+        const tableName =
+          modelName.toLowerCase() === 'planterregistration'
+            ? 'planter_registrations'
+            : modelName.toLowerCase();
+        const newQueryFields = whereObjClause.sql
+          .replace(/"/, `"${tableName}.`)
+          .replace(/AND "/gi, `AND "${tableName}.`)
+          .replace(/"/g, '');
+
+        whereObjClause.sql = newQueryFields;
+
         if (whereObjClause.sql) {
           const hasWhere = /WHERE(?![^(]*\))/i.test(selectStmt);
           query.sql += ` ${hasWhere ? 'AND' : 'WHERE'} ${whereObjClause.sql}`;
--- a/src/mixins/utils.repository-mixin.ts
+++ b/src/mixins/utils.repository-mixin.ts
@@ -1,14 +1,14 @@
 import { MixinTarget } from '@loopback/core';
 import { CrudRepository, Model } from '@loopback/repository';
 import expect from 'expect-runtime';
-import { buildFilterQuery } from '../js/buildFilterQuery';
-import { utils } from '../js/utils';
 
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 export function UtilsRepositoryMixin<
   M extends Model,
   R extends MixinTarget<CrudRepository<M>>,
 >(superClass: R) {
   return class extends superClass {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     [x: string]: any;
     // put the shared code here
     async getEntityIdsByOrganizationId(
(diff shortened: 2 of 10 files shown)


gwynndp:
@nmcharlton Do you want me to write tests for the mixin or not to bother because we're moving to microservices?

I'm also not sure about the queries for "Not set" organizations.  The queries filtered for null orgs and for any planters with null orgs.  I tried working on this because the captures page was always getting the same result as "All" for "Not set", but the numbers aren't adding up.

src/mixins/utils.repository-mixin.ts
```diff
@@ -0,0 +1,122 @@
+import { MixinTarget } from '@loopback/core';
+import { CrudRepository, Model } from '@loopback/repository';
+import expect from 'expect-runtime';
+import { buildFilterQuery } from '../js/buildFilterQuery';
```
nmcharlton: Can you try to clear down the linter warnings, or supress them where appropriate?

src/mixins/utils.repository-mixin.ts
```diff
@@ -0,0 +1,122 @@
...
+        const planterIds = await this.getNonOrganizationPlanterIds();
+        // if planter repository request
+        if (model === 'planter') {
+          return {
+            or: [{ organizationId: null }, { id: { inq: planterIds } }],
+          };
+        } else {
+          // if trees or other repository request
+          return {
+            or: [
```
nmcharlton: I think this needs to be `and`.
Only those captures from a non-org planter with no org attached should come up in the _Not set_ filter.

There are some planters not formally attached to an org (`planter.organizationId === null`) that plant on behalf of several orgs and might have those orgs attached to the captures. Those captures would be included in the `or` filter, but not `and`.

src/mixins/utils.repository-mixin.ts
```diff
@@ -0,0 +1,122 @@
...
+    async getOrganizationWhereClause(
+      organizationId: number,
+      model: string,
+    ): Promise<Object> {
+      if (organizationId === null) {
+        const planterIds = await this.getNonOrganizationPlanterIds();
+        // if planter repository request
+        if (model === 'planter') {
+          return {
+            or: [{ organizationId: null }, { id: { inq: planterIds } }],
```
nmcharlton: Both of the terms in this query equate to the same thing. `planterIds` is the set of planters where `organizationId === null`.

I'd suggest only doing the `getNonOrganizationPlanterIds()` lookup if `model !== 'planter'`.

nmcharlton (changes requested):
I'm inclined to say don't worry about tests, but this logic has caused us a few problems.
If basic tests can be added reasonably easily, it might prevent some headaches should we need to extend the functionality.

src/mixins/utils.repository-mixin.ts
gwynndp: Thank you, I'll try that

src/mixins/utils.repository-mixin.ts
gwynndp: Thank you! That makes the numbers add up

nmcharlton (approved):

nmcharlton (approved):

src/js/buildFilterQuery.js
```diff
@@ -28,6 +28,20 @@ export function buildFilterQuery(selectStmt, params) {
...
+          modelName.toLowerCase() === 'planterregistration'
+            ? 'planter_registrations'
+            : modelName.toLowerCase();
+        const newQueryFields = whereObjClause.sql
+          .replace(/"/, `"${tableName}.`)
+          .replace(/AND "/gi, `AND "${tableName}.`)
+          .replace(/"/g, '');
+
+        whereObjClause.sql = newQueryFields;
+
```
gwynndp, line 44: @nmcharlton 
I added this regex to add the modelName back onto each field because our more complex cross-table queries were resulting in "ambiguous" field errors that I couldn't find any other way of resolving.  Is there a better way to resolve those errors?

nmcharlton: I wonder whether the LoopBack query builders have a way of including the table name in the query.
This looks ok, but won't catch every `WHERE` clause construction.

PR #604, opened 2021-11-17
feat(capture-match): add fake data for capture match and fake endpoint

gwynndp:
DRAFT
- add mock api for capture-match tool


nmcharlton (commented):
Is there a reason for creating this endpoint here instead of in the Treetracker API?

PR #605, opened 2021-11-19
Fix query for country/region info for GrowerDetail

### LINKED ISSUE #122: Ability to set the country when editing a planter
Use case: During the verification process we often check the location of the tree and know the country. A simple drop down country list to assign the country to the planter would be a great addition. Same dialogue as the organization selector (edit planter)

gwynndp:
The SQL broke when adding the device-identifier to the query.
Partially Closes: https://github.com/Greenstand/treetracker-admin-client/issues/122

Both device-identifier and country are now working in the query
Growers - Treetracker Admin by Greenstand 2021-11-18 17-51-09

### DIFF
--- a/src/controllers/planterRegistration.controller.ts
+++ b/src/controllers/planterRegistration.controller.ts
@@ -31,7 +31,7 @@ export class PlanterRegistrationController {
   ): Promise<PlanterRegistration[]> {
     // console.log('/planter-registration', filter ? filter.where : null);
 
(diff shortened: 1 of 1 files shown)


nmcharlton (commented):
@gwynndp Did you mean to include the stakeholder changes in this PR?

gwynndp:
> @gwynndp Did you mean to include the stakeholder changes in this PR?

Nope, that wasn't intended. Sorry! I removed that commit now.

nmcharlton (approved):

PR #611, opened 2022-01-04
fix: add DISTINCT clause to ensure no duplicates

### LINKED ISSUE #609: Filtering trees by Any Tag Set returns duplicate records
Passing the following query to GET /trees returns duplicate records:
```
filter: "where":{"approved":true,"active":true,"tagId":"0"}
```

Trees (captures) appear to be included once per tag that they have applied, e.g., a tree with three tags appears three times in the results.

bstetzer32:
Added DISTINCT clause to SQL statements to ensure that trees with more than one tag do not result in duplicate entries.


nmcharlton (approved):
Thanks @bstetzer32

PR #645, opened 2022-08-13
feat(species):capture count if user has permission

### LINKED ISSUE #607: Species: Only query capture count if user has permission to manage species
We currently augment every species list query with the capture counts for each species:
https://github.com/Greenstand/treetracker-admin-api/blob/a400e8f2fa894de5d1bc11dec545e2e37037d578/src/controllers/species.controller.ts#L68

This is much more efficient than having the client query for each species individually, but adds several seconds onto the initial query.

Currently, the Species tool is only available to users with the `super_permission` policy in their roles (although soon that will include those with `manage_species`), which is the only place the counts are displayed in the client.

Therefore, there's no need to include these counts when the user doesn't have one of those policies.
That would significantly reduce the loading time for Organization users in particular.

kothariroshni8:
Resolves #607

### DIFF
--- a/src/controllers/species.controller.ts
+++ b/src/controllers/species.controller.ts
@@ -20,7 +20,7 @@ import {
 import { Request, RestBindings } from '@loopback/rest';
 import { Species } from '../models';
 import { SpeciesRepository, TreesRepository } from '../repositories';
-const allowedRoles = ['Admin', 'Species Manager'];
+const ALLOWED_POLICIES = ['super_permission'];
 
 export class SpeciesController {
   constructor(
@@ -71,12 +71,13 @@ export class SpeciesController {
     const speciesList = await this.speciesRepository.find(filter);
 
     if (this.requestHasUser(this.request)) {
-      //CheckUserAccess
-      const userRoles: string[] = this.request.user.roleNames;
-      const isAdminUser = userRoles.some((userRole) =>
-        allowedRoles.some((allowedRole) => allowedRole == userRole),
+      const userPolicies: any[] = this.request.user.policy.policies;
+      const isAllowedUser = userPolicies.some((userPolicy) =>
+        ALLOWED_POLICIES.some(
+          (allowedPolicy) => allowedPolicy == userPolicy.name,
+        ),
       );
-      if (isAdminUser) {
+      if (isAllowedUser) {
         return this.captureCountForEachSpecies(speciesList);
       }
     }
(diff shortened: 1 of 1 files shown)


src/controllers/species.controller.ts
```diff
@@ -61,23 +66,40 @@ export class SpeciesController {
...
-
+    const captureCounts = await this.treesRepository.execute(
+      captureCountStmt,
+      [],
+    );
     // Merge the capture counts with the species list
     return speciesList.map((species) => {
       const match = captureCounts.find((item) => {
-        return item.species_id === species.id;
+        return item.species_id == species.id;
```
nmcharlton: What's the reason for reducing this from `===` to `==`?

src/controllers/species.controller.ts
```diff
@@ -61,23 +66,40 @@ export class SpeciesController {
...
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ): Promise<any[]> {
-    console.log('get /species filter --> ', filter ? filter.where : null);
     // Only include active species
     filter = { ...filter, where: { ...filter?.where, active: true } };
+    const speciesList = await this.speciesRepository.find(filter);
 
+    if (this.requestHasUser(this.request)) {
+      //CheckUserAccess
+      const userRoles: string[] = this.request.user.roleNames;
```
nmcharlton: We should be checking user _policies_, not _roles_ (`this.request.user.policy.policies`).
See `src/js/auth.js` for the supported policies and how these checks are implemented (`isAuth()`).

nmcharlton (changes requested):
Sorry for the delay in reviewing your changes.
Great work so far, just a couple of comments.

src/controllers/species.controller.ts
kothariroshni8: I have updated it back to `===`. I recheck and update the same.

src/controllers/species.controller.ts
kothariroshni8: There were no species manager policy in Auth.js.
below are list of policies. which should i allow?

` SUPER_PERMISSION: 'super_permission',
  LIST_USER: 'list_user',
  MANAGER_USER: 'manager_user',
  LIST_TREE: 'list_tree',
  APPROVE_TREE: 'approve_tree',
  LIST_PLANTER: 'list_planter',
  MANAGE_PLANTER: 'manage_planter',`

nmcharlton: Just `SUPER_PERMISSION` is fine. We're not going to add the species manager policy to the legacy API.

kothariroshni8: Done.

nmcharlton (approved):
Nice work!

PR #667, opened 2026-06-02
feat: add greenstand-admin role in keycloak

Mohmn:


Mohmn:
fe pr @dadiorchen https://github.com/Greenstand/treetracker-admin-client/pull/1202

PR #670, opened 2026-06-12
fix: update type to 'O' from 'o' in org repo

Mohmn:


Mohmn:
@dadiorchen @mahdi2ba this needs to be merged before we go ahead witht the deployment

